PR #10, opened 2026-07-11
feat(agent-composer): add keyboard navigation to file mentions

### LINKED ISSUE #6: feat(agent-composer): add keyboard navigation to file mentions
## Problem

The Agent composer’s `@` file mention picker inserts the first matching file on Enter/Tab, without an active-result state or keyboard navigation. Selecting any result other than the first is therefore slower and less predictable.

## Screenshot

Image

## Proposed change

Make the file mention picker behave like a standard command/search palette:

- Focus the first matching result when the picker opens or its query changes.
- Use Arrow Up / Arrow Down to move the focused result, wrapping at the beginning/end.
- Use Enter and Tab to insert the focused file path into the composer.
- Keep the focused item visible in scrollable result lists.
- Sync pointer hover with the focused result.
- Add appropriate combobox/listbox semantics and a visible focus state.
- Do not add shortcut instructions; these interactions should remain implicit.

## Design

Preserve the current compact picker with small Codex-like refinements:

- File icon
- Filename as primary text
- Full relative path as muted secondary text
- Subtle focused-row background using existing design tokens

## Acceptance criteria

- Typing `@` shows matching library files.
- The first match is active by default.
- Arrow keys move the active result without moving the textarea caret.
- Enter and Tab add the active result’s path to the prompt.
- Escape dismisses the picker.
- Mouse selection continues to work.
- Keyboard and assistive-technology focus accurately reflect the active result.
- Existing composer behavior remains unchanged outside the mention picker.

PPRAMANIK62:
## Summary

- Add a CodeMirror-backed mention composer with atomic file tokens.
- Add Arrow Up/Down navigation with wrapping, Enter/Tab acceptance, Escape dismissal, pointer sync, and listbox/combobox semantics.
- Keep selected mentions intact when typing after them and insert a separating space automatically.
- Refine the compact picker styling with file icons, active-row state, and scrollable results.

## Screen Recording

https://github.com/user-attachments/assets/e17a8851-6529-4e75-ad79-991821160b64

## Root cause fixed

Mention markers were represented by a one-character document token. CodeMirror's default range mapping expanded the token when text was inserted immediately after it, causing marker validation to remove the mention. The marker now uses a left-biased end boundary.

## Validation

- `npx tsc --noEmit`

Closes #6

### DIFF
--- a/web-src/src/components/agent/AgentComposer.tsx
+++ b/web-src/src/components/agent/AgentComposer.tsx
@@ -1,10 +1,11 @@
-import { useEffect, useMemo, useRef, useState } from 'react';
+import { useEffect, useId, useMemo, useRef, useState } from 'react';
 import {
   ArrowUpIcon, BoltIcon, CheckIcon, ChevronDownIcon, ClipboardListIcon, CodeIcon, DumbbellIcon,
   FileGenericIcon, HandIcon, PlusIcon,
 } from '../../icons';
 import { useApp } from '../../store/AppContext';
 import { baseName } from './attachments';
+import { MentionComposer, type MentionComposerHandle, type MentionQuery } from './MentionComposer';
 import type { Attachment, EffortLevel, PermMode } from './types';
 
 const MODES: { id: PermMode; label: string; desc: string; Icon: typeof HandIcon }[] = [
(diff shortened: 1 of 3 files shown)


liliu-z (approved):
Looks good to me. I reviewed the CodeMirror mention composer path, keyboard navigation, serialization back to @path text, and local validation passes. The implementation is a bit heavier than the original picker, but the behavior and scope look reasonable.

PR #11, opened 2026-07-11
fix(agent-pane): smooth chat sidebar resizing

### LINKED ISSUE #7: fix(agent-pane): eliminate lag while resizing chat sidebar
## Problem

Dragging the right-hand Agent chat sidebar is visibly laggy, while resizing the file-explorer sidebar tracks the cursor smoothly.

The two resize paths diverge in two important ways:

- `.app` has a `220ms` `grid-template-columns` transition. The file-sidebar splitter applies `sidebar-dragging` during a gesture to disable that transition; the chat splitter does not. Each pointer move therefore restarts/interpolates the grid transition and makes the chat edge trail the cursor.
- The chat splitter dispatches `CHAT_WIDTH` through the global `AppContext` on every pointer move. This changes the provider value and rerenders mounted Agent views, including message transcripts, Markdown, and tool cards, while their width is changing.

The stylesheet comment claiming chat resize updates the layout “without re-rendering the heavy panes” does not match the current global-context implementation.

## Proposed change

Treat chat resizing as transient visual state until the gesture finishes.

- Add a chat drag class, mirroring `sidebar-dragging`, that disables the grid transition while the pointer is captured.
- During `pointermove`, calculate and clamp the prospective width, then update `--chat-width` directly on the `.app` element. Batch DOM writes with `requestAnimationFrame`.
- Do not dispatch `CHAT_WIDTH` on each pointer move.
- On `pointerup` and `pointercancel`, flush the final CSS-variable value, dispatch one `CHAT_WIDTH` action to persist it, and remove the drag class.
- Extract shared chat minimum/maximum width constants so drag-time clamping and the reducer use the same bounds.
- Update the stale chat-resize comment in `web-src/src/styles/chat.css`.

## Acceptance criteria

- Dragging the Agent chat splitter tracks the pointer without a visible 220ms lag.
- The grid transition is disabled only for the active chat-resize gesture and remains available for normal open/close behavior.
- Pointer movement does not dispatch `CHAT_WIDTH` at drag frequency.
- The final clamped chat width persists after releasing or cancelling the gesture.
- Resizing remains correct when the cursor leaves the splitter after pointer capture.
- The file-sidebar resize behavior remains unchanged.
- Renderer type-check and build pass:
  - `npx tsc --noEmit`

## Out of scope

- Virtualizing or otherwise redesigning the chat transcript.
- Changing chat tabs, agent transport, or session lifecycle.
- Altering normal sidebar open/close animation timing.

PPRAMANIK62:
## Summary

- Make Agent-chat resizing transient visual state during a drag.
- Batch direct `--chat-width` updates with `requestAnimationFrame` and persist the final width once.
- Disable grid transitions only during chat-resize gestures.
- Share chat-width clamping between the splitter and reducer.

## Screen Recording

https://github.com/user-attachments/assets/3553d6ac-a3aa-4cc0-a583-b1e83a99bcf2

## Root cause

The chat splitter dispatched `CHAT_WIDTH` through the global app context on every pointer move while the app's 220ms grid-column transition remained active. That repeatedly re-rendered mounted Agent views and made the panel edge trail the cursor.

## Validation

- `npx tsc --noEmit`
- `pnpm test:python`

Closes #7

### DIFF
--- a/web-src/src/App.tsx
+++ b/web-src/src/App.tsx
@@ -31,8 +31,7 @@ import { useHoverTip } from './hooks/useHoverTip';
 import { ErrorBoundary } from './components/ErrorBoundary';
 import { AppProvider, useApp } from './store/AppContext';
 import {
-  CHAT_MAX_WIDTH,
-  CHAT_MIN_WIDTH,
+  clampChatWidth,
   SIDEBAR_COLLAPSE_AT,
   SIDEBAR_MIN_WIDTH,
   SIDEBAR_MAX_WIDTH,
@@ -386,13 +385,9 @@ function ChatSplitter() {
   const pendingWidthRef = useRef<number | null>(null);
   const frameRef = useRef<number | null>(null);
 
-  function clampWidth(width: number) {
-    return Math.max(CHAT_MIN_WIDTH, Math.min(width, CHAT_MAX_WIDTH));
-  }
-
   function widthAt(e: ReactPointerEvent<HTMLDivElement>) {
     const start = startRef.current;
-    return start ? clampWidth(start.w - (e.clientX - start.x)) : null;
+    return start ? clampChatWidth(start.w - (e.clientX - start.x)) : null;
   }
 
   function writePendingWidth() {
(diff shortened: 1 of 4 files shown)


liliu-z (approved):
Looks good to me. I reviewed the transient CSS-var resize flow, single final CHAT_WIDTH dispatch, cleanup path, and local validation passes. Note: the 640px max-width cap is a product choice beyond the lag fix itself, but it fits the intended VS Code-like side-panel direction.

PR #15, opened 2026-07-14
feat(agent): improve Codex tool activity history

PPRAMANIK62:
## Summary

- restore missing desktop Codex tool calls from local rollout history without duplicating app-server items
- show compact, inspectable command/read/search activity labels and retain failure visibility
- keep restored tool activity collapsed by default and document the transcript behavior

## Screen Recording

https://github.com/user-attachments/assets/a069d2b6-284a-456e-924b-99fa4b07f184

## Why

Codex desktop histories can omit tool calls from the app-server transcript, and generic lifecycle summaries make completed activity hard to understand. This restores the missing history while keeping the transcript compact and actionable.

## Validation

- `pnpm exec tsc --noEmit`
- `pnpm exec vite build --config web-src/vite.config.ts`


liliu-z (approved):
Looks good to me — this is a nice improvement to the transcript. I especially like that the activity summary now describes what actually happened, like reads/searches/commands, instead of only showing lifecycle status.

The rollout-history fallback is scoped carefully to local Codex session files and avoids duplicating app-server tool items, so this feels reasonable to merge. Thanks for tightening up the history view here.

PR #24, opened 2026-07-15
feat(markdown-preview): add accessible Markdown footnotes

### LINKED ISSUE #20: feat(markdown-preview): add accessible Markdown footnotes
## Problem

The baseline preview has no footnote support, so technical notes and long-form documents lose citations and explanatory context. A minimal implementation must remain predictable for repeated references and usable with keyboard and assistive technology.

## What to build

Let readers open documents containing named footnotes, follow each superscript reference to one semantic footnote section, and navigate back to the exact originating reference—even where a footnote is cited repeatedly. Keep footnote parser behavior scoped to the isolated document renderer established by #18 so inline/chat rendering is unaffected, while preserving its conventional CommonMark-style soft-break and hard-break behavior.

## Acceptance criteria

- [ ] Named footnote definitions and references render in document preview.
- [ ] Footnote parsing is configured only for document preview and does not change inline/chat rendering.
- [ ] The document-preview soft-break and hard-break behavior established by #18 remains unchanged.
- [ ] References use semantic superscript links with descriptive accessible labels.
- [ ] Footnotes have stable entry IDs and backlinks to their originating references.
- [ ] Repeated references create unique reference targets without duplicating footnote content.
- [ ] Footnote typography, separators, focus states, and navigation are readable and keyboard-accessible.
- [ ] Tests cover named, repeated, malformed/edge-case footnotes, and isolation from inline rendering.

## Blocked by

#18

PPRAMANIK62:
## Summary

- add document-only named footnote definitions and references
- provide stable entry IDs, unique repeated-reference targets, accessible labels, and backlinks
- preserve CommonMark-style document breaks and keep inline/chat rendering isolated
- retain iframe fragment state so footnote targets and keyboard focus styles remain visible
- expand renderer coverage for footnote edge cases, sanitization, navigation, and baseline behavior

## Screen Recording

https://github.com/user-attachments/assets/8e4d52a1-bca2-4078-8bbd-231c3ca2a6f0

## Why the renderer is split

Issue #20 describes the footnote behavior, but it does not call out the supporting refactor. The previous `markdown.ts` combined parsing, sanitization, heading IDs, preview-document CSS, and orchestration in one module.

This PR splits those responsibilities into focused modules so the footnote extension remains document-only and cannot leak into inline Agent rendering. The same seams also make future Markdown preview features easier to extend and test without growing another monolithic renderer or weakening the preview trust boundary.

## User impact

Markdown previews now render named footnotes as a readable semantic section. Readers can follow repeated citations to one definition and navigate back to each exact originating reference using mouse or keyboard.

## Validation

- `pnpm test:renderer` — 26 tests passed
- `npx tsc --noEmit`
- `git diff --check main...HEAD`

Closes #20


liliu-z:
Thanks, this looks good to me. I like that the footnote behavior stays document-only and that the renderer split keeps inline/chat Markdown isolated from the preview pipeline.

The implementation is narrow enough to be reasonable here, and the test coverage around repeated references, malformed cases, fenced/raw HTML boundaries, sanitization, navigation, and focus styles is very helpful.

One thing I would keep in mind for the future: if the Markdown extension surface keeps growing, we may want to revisit whether this should move to a well-maintained Marked extension or plugin rather than maintaining more parser behavior ourselves. For this scoped footnote support, though, I think this is a good shape.

liliu-z (approved):
Approved. The footnote support is well-scoped to document preview, keeps inline/chat rendering isolated, and has strong coverage for navigation, repeated references, malformed cases, and sanitization boundaries.

Issue #17, opened 2026-07-15
Streamable HTTP transport for the MCP server (URL-only clients)


FenjuFu:
## Problem

The MCP surface is stdio-only today: clients spawn `~/.stashbase/bin/stashbase-mcp`, and the shim forwards every tool call over HTTP to the app server on `127.0.0.1:8090` (`mcp/server.ts` → `/api/library/*`). That works great for Claude Code / Codex / Claude Desktop, but a growing class of MCP clients can only attach to a **URL** (Streamable HTTP / SSE) and cannot spawn a local process:

- self-hosted agent platforms running in Docker on the same machine (e.g. [iFlytek astron-agent](https://github.com/iflytek/astron-agent) consumes MCP tools strictly by `mcp_server_url`; its workflow engine has a dedicated MCP node)
- web-based or sandboxed clients that expose "add MCP server by URL" only

For those users the library is unreachable even though the tool logic already sits behind HTTP.

## Proposal

Mount a Streamable HTTP transport on the existing Express server, e.g. `POST /mcp`:

- Reuse the exact tool surface of `mcp/server.ts` (`search_library`, `read_file`, `reindex`, `library_info`, bounded file helpers). Since every stdio tool already proxies to `/api/library/*`, the HTTP transport can share the same handlers — no second implementation path, staying consistent with the "one library, one MCP server" boundary in `design-docs/architecture.md` §7.2.
- Keep the bind on the IPv4 loopback as today, so nothing is exposed off-machine by default. Docker clients on the same host reach it via `host.docker.internal`.
- Settings → MCP gains a copyable URL next to the existing pasteable stdio config. Optional hardening: a locally generated bearer token shown in Settings, required on `/mcp` only.

## Why now

This is transport plumbing, not new capability: it turns "any MCP client that can spawn a process" into "any MCP client", which matches the README's `Local files -> ... -> MCP -> Agents` framing. Happy to implement it (including the docs update per the repo contract) if the direction sounds right — also fine to adjust scope, e.g. ship URL-without-token first or gate the whole endpoint behind a Settings toggle.

liliu-z:
Thanks, this direction makes sense to me. A URL-based MCP transport would make StashBase usable from clients that cannot spawn a local stdio process, while still exposing the same library/tool surface we already have.

A couple of design points I would want to be careful about before implementation:

- Keep this as transport plumbing only: the HTTP transport should share the same tool definitions and handlers as the stdio MCP server, rather than creating a second implementation path.
[...]

PR #25, opened 2026-07-16
feat(mcp): add token-gated Streamable HTTP transport for URL-only clients

FenjuFu:
Implements #17 along the lines you outlined — thanks for the quick direction check. This is the smallest version that proves the transport boundary.

## What

- **`mcp/library-server.ts`** (new): the tool definitions and handlers from `mcp/server.ts`, moved verbatim into a `createLibraryMcpServer({ webBase, windowId? })` factory. Transport-free by design — no stdio-guard, no Express, no argv/env reads.
- **`mcp/server.ts`**: now just the stdio entry — arg parsing, stdio-guard, `StdioServerTransport`. Behavior unchanged.
- **`server/routes/mcp-http.ts`** (new): `POST /mcp` on the app server. Stateless Streamable HTTP — each request builds a server instance from the shared factory and tears it down on response close; GET/DELETE answer 405 (no sessions or server-initiated streams to resume).
- **`server/index.ts`**: one mount line.
- **`design-docs/architecture.md`** §7.2/§7.3 + **README**: transports and auth model documented, including the Docker client shape (`http://host.docker.internal:8090/mcp`).

## Your design points, addressed

- **Transport plumbing only** — one implementation path: both transports build from the same factory, and every handler still forwards to the same `/api/library/*` routes.
- **Local by default** — binding untouched (`127.0.0.1`); nothing new listens anywhere.
- **Auth from the start** — bearer token required on every `/mcp` request, constant-time compare. Generated at boot into `~/.stashbase/mcp-http-token` (0600) so the file exists before the first client is configured. Rationale in §7.3: a spawned stdio server is reachable only by its spawner, a loopback URL by any local process — the token file restores "possession = local user". Browser-page vectors are additionally covered by the existing Origin allowlist (JSON POSTs from a rebound page carry a foreign `Origin` and get 403 before reaching the route).
- **Documented client shape** — README gains a "URL-based clients" subsection with the loopback and `host.docker.internal` forms.

Surfacing the URL + token in Settings → MCP is left as a follow-up to keep this reviewable; happy to do it next if you want it in the panel.

## Verification

- `tsc --noEmit` clean; `build:server` and `build:mcp` both bundle (esbuild follows the new module).
- stdio regression: piped `initialize` + `tools/list` into `mcp/server.ts` — same responses as before the refactor.
- HTTP end-to-end against a mounted Express app with a stubbed `/api/library/info`:
  - no token → 401 with a pointer to the token file; wrong token → 401
  - `initialize` → 200, `serverInfo: stashbase`
  - `tools/list` → all 9 tools
  - `tools/call library_info` → 200, forwarded through the shared handlers
  - `GET /mcp` → 405


liliu-z (approved):
Thanks, this looks good overall. I really like that this keeps the stdio and HTTP transports on the same MCP implementation path, and the token gate is a good starting boundary.

One small note: the Docker example is probably Docker Desktop-oriented for now. Since StashBase still binds to `127.0.0.1`, `host.docker.internal` may not work reliably on Linux Docker without a bit more design around networking. No need to change it in this PR though — I’ll merge this now, and we’ll clean up/document that follow-up on our side.

PR #37, opened 2026-07-17
refactor(agent): expand the Shared Agent Contract

### LINKED ISSUE #32: refactor(agent): expand the Shared Agent Contract
## Problem

The built-in panel presents Claude and Codex as one product surface, but the two runtime bridges currently carry different lifecycle, permission, history, and capability assumptions. That makes the renderer accumulate agent-specific branches and makes native CLI upgrades harder to absorb safely.

## What to build

Introduce a compatibility-first Shared Agent Contract for the built-in panel. It defines the common session lifecycle, normalized events, approvals, history actions, capability discovery, and adapter seam while preserving the existing Claude and Codex experience during migration.

## Design constraints

- The contract represents the panel promise, not a lowest-common-denominator runtime.
- Native-only behavior is exposed only through declared capabilities.
- The user's installed CLI remains the runtime and configuration source.
- The compatibility layer must allow incremental migration; no big-bang rewrite.

## Acceptance criteria

- [ ] One shared contract covers connection lifecycle, prompts, interruption, normalized transcript events, approvals, sessions, and capability discovery.
- [ ] Claude and Codex compatibility adapters preserve current panel behavior while routing through that contract.
- [ ] The contract has explicit states for available, unavailable, and failed runtimes.
- [ ] Capability discovery reflects the installed native runtime rather than a hard-coded version assumption.
- [ ] The renderer can consume the shared contract without adding new runtime-specific branches.
- [ ] Type-check and relevant build checks pass.

## Out of scope

- Changing the user's global Claude or Codex configuration.
- Changing the StashBase MCP server or library model.
- Altering the visible workflow before the adapter migrations land.

## Blocked by

None — can start immediately.

PPRAMANIK62:
## Summary

- introduce a compatibility-first Shared Agent Contract for Claude and Codex panel adapters
- route runtime discovery, websocket attachment, lifecycle cleanup, and session history through the contract
- move the renderer to capability-driven behavior and surface available, unavailable, and failed runtime states
- refresh runtime discovery after bridge errors so Codex app-server failures are visible and retryable

## Why

The built-in panel presented one product surface while its runtime bridges exposed separate lifecycle, history, and capability assumptions. The contract preserves the existing native bridges while making incremental migration safe.

## User impact

Claude and Codex retain their existing panel workflows. Runtime availability and failure state now reflect the installed CLI, and native-only behavior is declared through capabilities rather than renderer-specific branches.

## Validation

- git diff --check
- npx tsc --noEmit
- npx vite build --config web-src/vite.config.ts

Closes #32


liliu-z:
Looks good to me. I like that this keeps Claude and Codex native behind a compatibility contract instead of introducing a new product model. One small follow-up we may want later is moving the shared wire types out of `server/` into a neutral shared module, but that does not need to block this PR.

PR #40, opened 2026-07-18
feat(markdown-preview): highlight fenced code and polish preview

### LINKED ISSUE #21: feat(markdown-preview): highlight fenced code and polish preview
## Problem

Fenced code is currently plain, and several common document elements become hard to read or use in narrow windows. The preview needs polished technical-document rendering without weakening its iframe isolation or regressing navigation, search, images, and drag-and-drop.

## What to build

Show readable syntax-highlighted fenced code through marked-highlight and Highlight.js using its common-language set, retain safe fallback rendering for unlabelled or unknown code, and make long documents handle narrow layouts and common Markdown elements cleanly. Build these changes on the isolated document renderer established by #18 and the composed package-extension seam in #26 so inline/chat Markdown remains unaffected.

## Acceptance criteria

- [ ] Explicitly labelled common-language fenced code is highlighted without enabling iframe scripts or requiring runtime network access.
- [ ] Unknown and absent language labels remain readable as unhighlighted code blocks and never throw.
- [ ] Highlight colors fit the existing light preview palette.
- [ ] Tables scroll rather than overflow on narrow windows; long URLs and inline code wrap safely; task lists and supported semantic HTML have polished styles and keyboard focus states.
- [ ] Highlighting and preview-polish changes remain scoped to document preview and do not change inline/chat rendering.
- [ ] Existing local assets, in-document/cross-file anchors, external-link routing, image lightbox, find navigation, search-result highlighting, file-drop forwarding, and read-only isolation are regression-verified.

## Blocked by

#26

jashkarangiya:
## Summary

- highlight explicitly labelled fenced code with a static Highlight.js pass (common language set) on the document parser only; the iframe stays scriptless and needs no network access
- unknown or unlabelled blocks fall back to plain escaped code, and `ignoreIllegals` + try/catch ensure hostile input can never break rendering
- sanitizer allows `span` classes restricted to the Highlight.js token vocabulary (`hljs-*`, v11 sub-scopes like `function_`); no new tags, no inline styles
- preview polish: wide tables scroll inside the reading column, long URLs and inline code wrap, task lists drop the double bullet+checkbox, `kbd` / `mark` / `abbr` / `details` get styles with keyboard-focus outlines
- Agent-message Markdown stays unhighlighted; `design-docs/markdown-rendering.md` updated in the same change

## Find fixes

- Find matched within single text nodes, so token spans from highlighting would have broken phrase queries inside code blocks. Matching now runs over the concatenated text of visible text nodes and maps back to multi-node `Range`s, the same approach the search-chunk highlighter already uses. `server/html.ts` keeps its per-node copy (content unaffected); porting it is a follow-up.
- Pre-existing: clicking a keyword hit for the already-open file showed 0/0 until the query was retyped, because the controller only ran on viewer mount or content reload. `selectFileWithHighlight()` now primes the live controller directly.

## Validation

- `pnpm test:renderer` (33 tests, including new highlighting fixtures: fallbacks, sanitizer smuggling, hostile labels, span class stripping, inline isolation)
- `npx tsc --noEmit`
- regression-checked live: anchors, image lightbox, keyword search into the preview, find across highlighted tokens, narrow-window layout, read-only isolation

Adds ~100 KB gzipped to the renderer bundle from the Highlight.js common language set.

### Before
before

### After
after

Closes #21


PPRAMANIK62:
@jashkarangiya This looks good to me. One small find regression to address before merge:

`collectMatches()` concatenates every text node without inserting block boundaries. Supported raw HTML like `<p>foo</p><p>bar</p>` therefore becomes searchable as foobar, producing a false match across two separate paragraphs.

Could we preserve joins across inline/highlight spans, but add a separator (or reset matching) when traversal crosses block-level elements? A regression test for this raw-HTML case would cover it.

jashkarangiya:
Good catch! Fixed in [9e8e6dc](https://github.com/liliu-z/stashbase/pull/40/commits/9e8e6dc64112723c99d539726ad46e068c33c65d).

Traversal now inserts newline separators at block-level boundaries and `<br>` elements, preventing adjacent raw HTML blocks from being merged into a single match, while still allowing inline and highlight spans to match as expected. Since a query can never contain a newline, matches can no longer span across those separators.

Also added find-corpus.test.ts to cover the `<p>foo</p><p>bar</p> `case, along with separation across `<br>`, list items, and table cells.

PPRAMANIK62 (approved):

PPRAMANIK62:
Thanks!! The changes and the recent fix added looks good to me !!

PR #56, opened 2026-07-20
feat(search): add subfolder scoping and file-type filters

### LINKED ISSUE #31: feat(search): add subfolder scoping and file-type filters
## Problem

Search over an open folder is all-or-nothing. The indexer and daemon already support prefix-scoped search (`path_prefix`), but only the MCP `search_library` tool can use it. The app's Search panel can't. There's also no way to narrow results by file type, so a library mixing PDFs, images, and notes always returns everything blended together.

The README lists "Search filters and ranking controls" as still evolving, and this proposes the filters half.

## What to build

- Let the Search panel scope a query to a subfolder, in both semantic and keyword modes. This mostly plumbs the existing `path_prefix` through `POST /api/search`.
- Add file-type filters (Markdown/HTML, PDF, images) that work in both modes.

Ranking controls (score display, top-k, grouping) stay out of scope here and can be a follow-up.

## Acceptance criteria

- [ ] A query can be scoped to a subfolder, and clearing the scope restores folder-wide search.
- [ ] File-type filters narrow results in both modes and compose with scoping.
- [ ] Derived notes still never surface, and hits still remap to their source PDF/image.

jashkarangiya:
The in-app Search panel could not narrow results the way the indexer and MCP path already allowed. Add two composable filters to both semantic and keyword modes:

- Subfolder scope: a folder-relative subfolder resolved escape-safe against the active folder (rejects traversal, missing, non-dir).
- File-type categories (notes / pdf / image / docx) mapped to source extensions in server/format.ts, sharing one vocabulary via shared/search-types.ts.

The semantic path passes the extension filter to the daemon, which over-fetches from MFS (bounded), filters hits by source suffix, and truncates back to top_k, so filtering happens before the caller-visible cut. The keyword path restricts the ripgrep target and derived-text walk to the scoped subtree and enabled categories. Display-path remapping is unchanged: filters act on source paths and derived notes never surface.

## Before / After

Same query, same keyword mode — the only difference is the new subfolder-scope dropdown and file-type chips.

| Before | After |
|---|---|
| search sidebar without filters | search sidebar with subfolder scope and file-type chips |

Closes #31

### DIFF
--- a/design-docs/architecture.md
+++ b/design-docs/architecture.md
@@ -98,7 +98,7 @@ daemon adapters cross these modules. Identity, containment, migration, and proto
 
 One installation has **one library**: the set of opened folders indexed into one collection and exposed by one MCP server.
 
-MCP search defaults to the whole library. Calls can narrow scope by folder root or path prefix. The in-app search UI is scoped to the current window's folder.
+MCP search defaults to the whole library. Calls can narrow scope by folder root or path prefix. The in-app search UI is scoped to the current window's folder and can narrow further to one subfolder and to file-type categories.
 
 On server boot, StashBase binds every library folder into the daemon and then reconciles them in the background. The Welcome screen also reconciles library folders with a short cooldown and polls folder status when it is idle. While a folder is actively opening, Welcome status polling and reconcile are deferred so navigation does not compete with preparation work.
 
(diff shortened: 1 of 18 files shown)


liliu-z:
Nice work on this — the feature is useful, and I like that it works consistently across semantic and keyword search.

Purbayan Pramanik is currently working on some broader UI refinements, so we may make a few small adjustments to the interaction/design later, but this looks good to merge.

PR #60, opened 2026-07-22
feat(markdown-editor): present fenced code with whole-block reveal

### LINKED ISSUE #50: feat(markdown-editor): present fenced code with whole-block reveal
## What to build

Present inactive fenced code as an inert monospace block while preserving the exact Markdown source. Entering or selecting anywhere inside the block reveals both fences and the language label as one editable construct.

Known language labels may drive presentation styling, but unsupported or absent labels must remain readable and must never throw. Live Editing does not execute code or insert user-authored HTML into the application DOM.

## Acceptance criteria

- [ ] Inactive fenced code has a stable monospace block presentation.
- [ ] Entering or selecting anywhere inside a code block reveals both fences and its language label.
- [ ] Leaving the block restores its inactive presentation without changing source or history.
- [ ] Labeled, unlabeled, and unknown-language blocks remain readable.
- [ ] Fence characters inside code content and incomplete fences do not corrupt block boundaries.
- [ ] Code is never executed and raw HTML is never promoted into a live application widget.
- [ ] Selection, copy/cut, Find, undo/redo, autosave, and scroll stability are regression-covered.
- [ ] Reading View fenced-code rendering remains independently authoritative.
- [ ] Current-state design documentation describes the shipped Live code-block behavior.
- [ ] Repository typecheck and renderer build pass.

## Blocked by

- #45

jashkarangiya:
Adds fenced code as a Live Editing construct so code blocks read as blocks instead of raw prose, following the existing construct and reveal pattern.

## What it does

- Inactive fenced code renders as an inert monospace surface with the fences and language label concealed.
- Entering or selecting anywhere inside the block reveals both fences and the language label as one editable construct, and leaving restores the block.
- The projection never changes source or history. A small line-class capability on the projection rule gives the block its full-width presentation.
- The parser owns block boundaries, so backticks inside content and unterminated fences never split a block, and the language label is only concealed markup that is never parsed or executed.
- Reading View fenced rendering stays independently authoritative.

## Before / After

| Before | After |
|---|---|
| fenced block as raw source | fenced block as an inert monospace block |

Before, a fenced block shows raw ``` fences in the body font and reads like prose. After, it is an inert monospace block with the fences and label concealed.

Entering the block reveals its fences and label as one editable construct, while other blocks stay concealed:

editing a fenced block reveals its fences

## Acceptance checklist

- [x] Inactive fenced code has a stable monospace block presentation.
- [x] Entering or selecting inside reveals both fences and its language label.
- [x] Leaving restores the inactive presentation without changing source or history.
- [x] Labeled, unlabeled, and unknown-language blocks remain readable.
- [x] Fence characters inside content and incomplete fences do not corrupt boundaries.
- [x] Code is never executed and raw HTML is never promoted into a live widget.
- [x] Selection, copy/cut, Find, undo/redo, autosave, and scroll stability are regression-covered.
- [x] Reading View fenced rendering remains independently authoritative.
- [x] Current-state design documentation updated.
- [x] Typecheck and renderer build pass.

## Verification

The renderer test suite passes, including new fenced-code tests for reveal and restore, labeled, unlabeled, and unknown languages, fenced content, unterminated fences, Find, and undo. Typecheck and renderer build pass. I also drove it in the app to confirm the block presentation, per-block reveal, and no console errors.

Closes #50

### DIFF
--- a/design-docs/design/markdown.md
+++ b/design-docs/design/markdown.md
@@ -14,6 +14,12 @@ without a conversion layer becoming the product.
   changing their source offsets: click reveals source, modifier-click or
   keyboard activation follows safe local links in StashBase and HTTP(S) links
   through the system browser.
+- In Live Editing, fenced code presents as an inert monospace block with its
+  fences and language label concealed; entering or selecting inside reveals
+  them as one editable construct and leaving restores the block, all without
+  changing source. The parser owns block boundaries, so backticks inside
+  content and unterminated fences do not corrupt them, and the language label
+  is only concealed markup — never parsed or executed.
 - Safe local links remain in StashBase; external links retain their normal
   browser behaviour.
 - Agent responses and Markdown documents remain distinct presentation contexts.
--- a/web-src/src/__tests__/code-editor.test.ts
+++ b/web-src/src/__tests__/code-editor.test.ts
@@ -86,6 +86,10 @@ test('Live Editing and Reading View share the supported Markdown construct subse
     '',
     '---',
     '',
+    '```ts',
+    'const value = 42;',
+    '```',
+    '',
   ].join('\n');
 
   const projectedKinds = new Set(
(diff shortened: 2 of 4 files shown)


liliu-z:
LGTM — this fits the Live Editing direction well. Thanks for the careful implementation and the regression coverage here.

PR #69, opened 2026-07-23
feat(search): surface relevance and add a result-count control for semantic search

### LINKED ISSUE #68: feat(search): surface relevance and add a result-count control for semantic search
## Problem

Semantic search returns ranked hits, but the Search panel shows a flat list with no sense of how many results there are, how strongly each one matches, or any control over how many to show. Each hit already carries a `score` (returned from the daemon and present on `SearchHit`), yet it is never surfaced. Keyword search already shows a result summary; semantic search does not.

This is the "ranking controls" half of the roadmap line "Search filters and ranking controls" (the filters half shipped in #56).

## What to build

- A result-count summary for semantic results, matching the existing keyword summary.
- Progressive disclosure: show the top results first with a control to reveal the rest of the already-fetched candidates, so a long list is not dumped at once. Clearing the query or switching folders resets it.
- A subtle per-hit relevance indicator showing relative match strength within the current result set. It is a visual cue derived from the existing ordering, not a raw score, so it stays honest even though hybrid RRF scores have no absolute meaning.

This is renderer-only: the score and candidate list already come back from the API, so no server or daemon change is needed.

## Out of scope

Changing the ranking algorithm, exposing raw score numbers, or adding re-ranking controls (sort by date, semantic vs keyword weighting). Those are a separate, larger effort.

## Acceptance criteria

- [ ] Semantic results show a result count.
- [ ] Only the top N show initially, with a control to reveal the rest; clearing the query or changing folders resets it.
- [ ] Each semantic hit shows a relative relevance indicator that degrades gracefully for a single result and for all-equal scores.
- [ ] Keyword mode is unchanged.
- [ ] Typecheck and renderer build pass.

jashkarangiya:
Makes semantic ranking legible in the Search panel. This is the ranking half of the roadmap line whose filters shipped in #56.

## What it does

Semantic results were a flat list: the hits are ranked, but nothing showed how strongly each one matched, even though every hit already carries a score. This adds:

- A relative match-strength bar per hit, right aligned in the hit meta row.
- A top-N summary line, matching the summary keyword search already has.

The bar is normalized within the shown result set, not a raw score, because hybrid RRF scores have no absolute meaning. Real scores for one query ranged from 0.0312 to 0.0180, so raw numbers would look precise and tell the user nothing, while the normalized bars read as 100, 96, 85, 52, 44, 34, 34, 20 percent. It stays honest for a single result and for all-equal scores, where every bar fills.

Renderer only. The score already comes back from the API, so there is no server or daemon change, and all new styles use existing theme tokens.

## What real-flow testing changed

I trialled this against a real semantic index (17 documents, roughly 300 chunks) rather than shipping it on assumption, and it corrected the original design:

- The first version added progressive disclosure with a show-more control. In practice `filterGuiSemanticHits` already narrows results to the strongest matches and hard caps them at `SEMANTIC_SEARCH_MAX_VISIBLE` (8) before they reach the panel, so that control could never render. The API returned 30 candidates and the panel received 8. It is removed rather than left as dead code.
- A plain "8 results" count implied only 8 documents matched, when weaker candidates had been dropped by that cutoff. The summary now reads "Top 8 matches", which is accurate about what the cutoff leaves.

Worth noting for a possible follow-up, entirely your call: the cutoff means users never see beyond the top 8 semantic matches, and there is currently no way to look further. That is a deliberate quality decision, so I did not touch it here.

## Before / After

Full app window, captured against the real index. Same query and same opened result in both, so the only difference is the search panel:

semantic results before and after: flat ranked list versus a top-N summary with per-hit match-strength bars

## Acceptance criteria

- [x] Semantic results show a summary line.
- [x] Each hit shows a relative relevance indicator that degrades gracefully for a single result and for all-equal scores.
- [x] Keyword mode is unchanged.
- [x] Typecheck and renderer build pass.

## Verification

Unit tests cover the relevance normalization (floor and range, order independence, single result, all-equal, empty, negative scores). The full renderer suite passes at 104 tests, and typecheck and renderer build pass. Driven end to end in the app against a real semantic index with no console errors.

Closes #68

### DIFF
--- a/design-docs/design/search.md
+++ b/design-docs/design/search.md
@@ -11,6 +11,10 @@ as the result identity.
 - In-app search starts from the current folder and can narrow its scope.
 - Results identify the source file, path, and useful evidence such as a snippet
   or page/timestamp hint.
+- Semantic results show a result count and a per-hit relative match-strength
+  indicator, and reveal a long candidate list progressively. The indicator is
+  relative to the current result set, not an absolute score, because hybrid
+  scores have no absolute meaning.
 - Prepared PDF, image, DOCX, and media transcript text can be evidence, but
   opening a result returns to the original source file.
 - MCP offers orientation, search with the same file-type categories as the
--- a/design-docs/design/search.md
+++ b/design-docs/design/search.md
@@ -11,11 +11,13 @@ as the result identity.
 - In-app search starts from the current folder and can narrow its scope.
 - Results identify the source file, path, and useful evidence such as a snippet
   or page/timestamp hint.
-- Semantic results are already narrowed to the strongest matches by a
-  relevance cutoff, and they are presented as a top-N summary with a per-hit
-  relative match-strength indicator. The indicator is relative to the shown
-  result set, not an absolute score, because hybrid scores have no absolute
-  meaning.
+- Semantic results show the strongest matches first, up to a relevance-knee
+  count, and reveal the remaining fetched candidates through progressive
+  disclosure without another request. The summary reports the visible and
+  available counts (for example, "Showing 8 of 30 ranked candidates"), and
+  each hit carries a per-hit relative match-strength indicator. The indicator
+  is relative to the fetched result set, not an absolute score, because hybrid
+  scores have no absolute meaning.
 - Prepared PDF, image, DOCX, and media transcript text can be evidence, but
   opening a result returns to the original source file.
 - MCP offers orientation, search with the same file-type categories as the
--- a/design-docs/design/search.md
+++ b/design-docs/design/search.md
@@ -11,10 +11,11 @@ as the result identity.
 - In-app search starts from the current folder and can narrow its scope.
 - Results identify the source file, path, and useful evidence such as a snippet
   or page/timestamp hint.
-- Semantic results show a result count and a per-hit relative match-strength
-  indicator, and reveal a long candidate list progressively. The indicator is
-  relative to the current result set, not an absolute score, because hybrid
-  scores have no absolute meaning.
+- Semantic results are already narrowed to the strongest matches by a
+  relevance cutoff, and they are presented as a top-N summary with a per-hit
+  relative match-strength indicator. The indicator is relative to the shown
+  result set, not an absolute score, because hybrid scores have no absolute
+  meaning.
 - Prepared PDF, image, DOCX, and media transcript text can be evidence, but
   opening a result returns to the original source file.
 - MCP offers orientation, search with the same file-type categories as the
--- a/web-src/src/__tests__/search-relevance.test.ts
+++ b/web-src/src/__tests__/search-relevance.test.ts
@@ -1,0 +1,35 @@
+import assert from 'node:assert/strict';
+import test from 'node:test';
+
+import { relevanceRatios } from '../lib/searchRelevance.ts';
+
+const approx = (actual: number, expected: number) => Math.abs(actual - expected) < 1e-9;
+
+test('relevance normalizes to a floored [0.2, 1] range within the result set', () => {
+  const ratios = relevanceRatios([1, 0.5, 0]);
+  assert.equal(ratios[0], 1); // strongest fills the bar
+  assert.equal(ratios[2], 0.2); // weakest keeps a visible floor
+  assert.ok(approx(ratios[1], 0.6)); // midpoint scales between floor and 1
+});
+
+test('relevance is order-independent and driven by value, not position', () => {
+  const ratios = relevanceRatios([0, 0.5, 1]);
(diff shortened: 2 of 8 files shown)


liliu-z:
Since this PR directly affects the actual search experience, could you try it with a real semantic search flow first and see how it feels in use? I’ll send you a test API key on Slack.

jashkarangiya:
Good call on trying it for real, it caught two things.

The show-more control could never fire. `filterGuiSemanticHits` already caps results at 8 before they reach the panel, so the API returns 30 candidates and the panel gets 8. Removed it. And "8 results" was misleading when weaker matches had already been cut, so it now reads "Top 8 matches".

The relevance bar held up well though. Raw scores were 0.031 down to 0.018, useless as numbers, but normalized they read 100/96/85/52/44/34/34/20%. Before/after in the description are real captures now.

One thing I left alone: with that cutoff you can never see past the top 8 matches. Looked like a deliberate call, so I didn't touch it.

Thanks for the key, only used it for one indexing pass and a few queries.

HonQii (changes requested):
Thanks for testing this against a real semantic index. The relative relevance indicator is well-contained, avoids exposing raw hybrid scores, and handles the single-result/all-equal cases cleanly.

I am requesting changes because the PR still says it closes #68 while omitting that issue's central progressive-disclosure requirement.

### [P1] Preserve the fetched candidates

The API already returns up to 30 candidates, but `useSearchActions` applies `filterGuiSemanticHits(hits)` before dispatch, so everything beyond the initial quality cutoff is discarded before the renderer receives it. Removing the unreachable control documents the current limitation; it does not satisfy #68.

A cleaner implementation would:

1. Preserve the complete fetched candidate list in search state.
2. Refactor the cutoff logic to derive an `initialVisibleCount` instead of deleting the remaining candidates.
3. Render the initial slice and add a control that reveals the remaining already-fetched results without another request.
4. Reset expansion when the query, folder, scope, type filter, or search mode changes.
5. Add component/state coverage for expansion and reset behavior.

Relevant code: `web-src/src/components/SearchPanel.tsx:394` and `web-src/src/store/useSearchActions.ts:427` on the simulated merge with current `main`.

### [P2] Report visible and available counts separately

`Top N matches` only reports the post-cutoff visible count, not the number of candidates returned by the search. Once the complete list is preserved, please use wording such as `Showing 8 of 30 ranked candidates`, which does not imply a corpus-wide total. The singular case should retain an explicit numeric count if the goal is to match keyword search's count feedback.

Relevant code: `web-src/src/components/SearchPanel.tsx:403`.

### [P2] Follow the current UI-scale typography contract

On the latest `main`, adjacent semantic-search typography and `.keyword-summary` use `calc(... * var(--ui-scale))`. The new `.search-summary` keeps a fixed `11px`, so it does not respond to the Small/Large appearance settings. Please use:

```css
font-size: calc(11px * var(--ui-scale));
```

I simulated the merge onto the current `main`; typecheck, the renderer build, and all 106 renderer tests pass. The blocker is specification completeness rather than an existing build or regression failure.

jashkarangiya:
Thanks @HonQii, this was the right call. Removing the control documented the limitation instead of delivering #68, and it is implemented properly now. I rebased onto current main first.

**P1 (preserve candidates + real disclosure).** `useSearchActions` now dispatches the full fetched candidate list instead of slicing it. The old `filterGuiSemanticHits` cutoff became `guiSemanticVisibleCount`, which returns the relevance-knee count. The panel renders that initial slice and a control reveals the remaining already-fetched candidates in steps with no new request. Disclosure resets whenever a fresh result set arrives, which covers query, folder, scope, type-filter, and mode changes, since each of those dispatches a new hits array.

**P2 (counts).** The summary now reads `Showing 8 of 30 ranked candidates`, and the single-result case keeps an explicit numeric count (`1 ranked candidate`).

**P2 (typography).** `.search-summary` and the new `.search-show-more` use `calc(... * var(--ui-scale))`, so they follow the Small/Large appearance settings.

Verification: typecheck, renderer build, and the renderer suite pass (124 tests, including new coverage for `guiSemanticVisibleCount`). I also drove a real semantic search against an indexed folder: it started at `Showing 8 of 30 ranked candidates` with a Show more control, one click revealed `Showing 16 of 30` from the already-fetched set, and a new query reset back to the initial slice, with no console errors. Since the repo has no component-test harness, expansion and reset are covered by the count-logic unit tests plus that live run.

HonQii (approved):
The requested progressive-disclosure, count, and UI-scale changes are now implemented. I rechecked the exact head against current main: the merge is clean, the renderer tests, both TypeScript checks, the production renderer build, and the chunk gate pass, and all three platform CI jobs are green. Approved for merge.

PR #95, opened 2026-08-01
fix(welcome): keep folder opens safe from stale index status

### LINKED ISSUE #89: fix(welcome): stale index-status 412 can discard a successful New Folder open
## Summary

Creating and selecting a folder from a fresh Welcome screen can succeed on disk and on the server while the renderer remains on Welcome with no error.

Observed on macOS from source at `eefdd28ff5ba` (`package.json` 1.3.2).

## Reproduction

1. Launch StashBase into the Welcome screen with no active folder.
2. Click **New**.
3. In the native folder picker, create a folder and confirm **Select folder**.
4. Return to the app.

## Actual behavior

- The folder is created on disk.
- The root `AGENTS.md` is created.
- The folder is added to the Library.
- The server binds the selected folder as the current folder for the renderer window, and `/api/files` works for it.
- The renderer nevertheless stays on Welcome indefinitely, without an error.

Clicking the new Library row again is a workaround.

## Expected behavior

Once `POST /api/folder` succeeds, the current window should immediately enter the selected folder view. File listing, ordering, reconciliation, and indexing may continue in the background.

This is required by the navigation contract in `code-review/data-layer.md` §9.1 and `code-review/architecture.md` §2.2.

## Likely root cause

There is a race between the shared folder-open generation and index-status polling:

1. `NewFolderButton` receives the selected native path and calls `actions.openFolder()` (`web-src/src/components/Welcome.tsx`).
2. `openFolder()` increments `openGeneration`, sends `POST /api/folder`, and later discards the successful response if that generation changed (`web-src/src/store/useFolderActions.ts`).
3. Active-folder index polling continues while the renderer is on Welcome. A request started with no folder can return `412 NO_FOLDER` after the newer open has begun (`web-src/src/store/useSearchActions.ts`).
4. Because the renderer `folderPath` is still empty until `finishOpenFolder()` runs, `stillTargetFolder()` treats that stale Welcome response as current. The 412 handler then increments the same `openGeneration` unconditionally.
5. The folder POST has already succeeded server-side, but its renderer continuation now sees a generation mismatch and returns before `finishOpenFolder()` / `WELCOME_HIDE`.

This leaves server and renderer folder contexts split and matches the observed state exactly.

## Acceptance criteria

- Creating/selecting a folder from a fresh Welcome screen always enters that folder view after the open API succeeds.
- The same guarantee holds for the shared Open/New/recent-folder action path.
- A stale `NO_FOLDER` / index-status response started before or during a newer open cannot advance or cancel that open generation.
- Preserve the intended invalidation behavior for explicit Home, folder removal, window retirement, and real server-restart recovery.
- Add a deterministic renderer regression test that:
  1. defers a Welcome `/api/index-status` 412,
  2. starts a folder open,
  3. resolves the stale 412 around the successful open response,
  4. asserts that the selected `folderPath` is committed and Welcome is hidden.
- Consider skipping active-folder index polling entirely while the renderer has no active folder; Welcome already owns explicit per-library-folder status polling.

## Existing test gap

The current renderer suite passes, but it covers the folder mutation queue and membership rebind helpers rather than this hook-level interleaving. Electron multi-window smoke opens the folder endpoint directly and therefore bypasses the React Welcome/open/status race.

swapnanilhoichoi:
## Summary

Fixes #89.

Opening a folder from Welcome could complete on the server while the UI remained on Welcome. A late `NO_FOLDER` response from an older index-status poll could invalidate the newer folder-open operation before the renderer committed the selected folder.

This change:

- skips active-folder status polling until the renderer has committed a folder
- fences status responses with both the folder path and navigation generation
- keeps recovery for a genuinely lost active-folder binding
- adds a regression test for the stale Welcome `412` interleaving

## Validation

- `pnpm test:renderer`
- `pnpm typecheck`
- `pnpm test:electron`
- `pnpm test:electron:smoke`

### DIFF
--- a/code-review/data-layer.md
+++ b/code-review/data-layer.md
@@ -433,7 +433,7 @@ Representative failure:
 Current contract:
 
 - Opening a folder establishes renderer folder identity and hides Welcome before `/api/files`, file order, or `/api/index-status` finish.
-- Active-folder index polling starts only after that renderer identity is committed. Every poll response, including `NO_FOLDER`, is fenced by both the committed folder path and the navigation generation, so a Welcome-era or superseded response cannot cancel a newer successful open.
+- Active-folder index polling starts only after that renderer identity is committed. Every poll response, including `NO_FOLDER`, is fenced by both the committed folder path and the navigation generation, so a Welcome-era or superseded response cannot cancel a newer successful open. Rebind recovery rechecks that fence after every await and shares the window's folder-mutation queue with user navigation, so an old recovery cannot restore an obsolete server binding.
 - Folder-switch side effects such as index sync and Agent-session cleanup are scheduled after the open-folder response, so they cannot turn navigation into a failed request.
 - Folder-sync queue cleanup consumes rejected promises; an indexer or daemon startup failure records an index warning but must not exit the Node server.
 - Going Home resets renderer folder state and shows Welcome before the server-side close request returns.
--- a/code-review/data-layer.md
+++ b/code-review/data-layer.md
@@ -433,6 +433,7 @@ Representative failure:
 Current contract:
 
 - Opening a folder establishes renderer folder identity and hides Welcome before `/api/files`, file order, or `/api/index-status` finish.
+- Active-folder index polling starts only after that renderer identity is committed. Every poll response, including `NO_FOLDER`, is fenced by both the committed folder path and the navigation generation, so a Welcome-era or superseded response cannot cancel a newer successful open.
 - Folder-switch side effects such as index sync and Agent-session cleanup are scheduled after the open-folder response, so they cannot turn navigation into a failed request.
 - Folder-sync queue cleanup consumes rejected promises; an indexer or daemon startup failure records an index warning but must not exit the Node server.
 - Going Home resets renderer folder state and shows Welcome before the server-side close request returns.
--- a/web-src/src/__tests__/folder-path.test.ts
+++ b/web-src/src/__tests__/folder-path.test.ts
@@ -46,3 +46,21 @@ test('context recovery never reopens a folder removed by another window', async 
   assert.equal(opens, 0);
   assert.equal(result.opened, null);
 });
+
+test('context recovery does not issue an old folder open after navigation changes during its library read', async () => {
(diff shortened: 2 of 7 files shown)


PPRAMANIK62 (changes requested):
Request changes

1. **P1 — stale 412 recovery is not re-fenced after an await** (`web-src/src/store/useSearchActions.ts:299-340`). The code verifies the poll generation before `await rebindFolderIfStillInLibrary(...)`, but not afterward. If a newer folder open starts during the rebind, this recovery can fall through to the hard reset, increment `openGen`, and return the renderer to Welcome after the newer folder has succeeded server-side. Re-check `stillTargetFolder()` / the generation after recovery and before any recovery dispatch or reset; ensure the stale rebind cannot overwrite a newer server binding.

2. **P2 — regression test does not exercise the required interleaving** (`web-src/src/store/__tests__/folder-open-status-race.test.ts:12-48`). It manually tests `isCurrentIndexStatusPoll` and the reducer, but does not run `refreshIndexState` with deferred `/api/index-status` 412 and `/api/folder` responses. Add a deterministic hook-level or integration test proving the actual open continuation commits the selected folder and hides Welcome when that stale response resolves.

These are required to satisfy #89: every stale status response must be unable to cancel a newer successful folder open.

swapnanilhoichoi:
Thanks — both points are addressed in b836b61.

1. The `412` recovery path now re-validates the poll ownership after the library read and again before any reset. Recovery and user-driven Open/Home transitions also share one per-window mutation queue. That means a recovery made stale during an await either never issues its old open request, or a newer user open is serialized after it and remains the final server binding.

2. I replaced the helper-only coverage with deterministic hook-level coverage. The test invokes the real status and folder-open actions with deferred API responses, verifies that a stale `412` cannot prevent the selected folder from committing or Welcome from hiding, and adds a separate case for a recovery that becomes stale during its library read.

Validation after the update:

- `pnpm test:renderer` (87 passing)
- `pnpm typecheck`
- `pnpm test:electron`
- `pnpm test:electron:smoke`

PPRAMANIK62:
Closing this PR because the underlying issue has already been fixed and merged in another PR. Thank you for the investigation and regression coverage.

PR #111, opened 2026-08-04
fix(packaging): harden installation recovery

swapnanilhoichoi:
## Summary
- clarify supported macOS, Windows, and Debian/Ubuntu installation lifecycle guidance
- preserve and restore the previous macOS app when recovery replacement fails
- exercise recovery rollback failures in source CI and macOS DMG release verification

## Note
- renderer typechecking remains blocked locally because node_modules is stale and missing dependencies declared by the merged frontend work.


build/dmg-scripts/Fix.sh
```diff
@@ -35,9 +35,17 @@ SOURCE_APP="$1"
 TARGET_APP="$2"
 SIGN_SCRIPT="$3"
 stage="starting"
+BACKUP_APP=""
 
 fail() {
   local code="$?"
+  trap - ERR
+  set +e
+  if [[ -n "$BACKUP_APP" && -d "$BACKUP_APP" ]]; then
```
HonQii: **[P1] Roll back on interruption and every uncommitted exit**

This restoration path only runs through `fail()`, while the helper traps only `ERR`. After the existing app has been moved to `*.stashbase-previous-*`, `SIGINT`, `SIGTERM`, or `SIGHUP` can terminate the helper without restoring it. I reproduced this by interrupting the signer: the new, unverified target remained in place and the old runnable app was left hidden in the backup directory. Please handle `EXIT`/`INT`/`TERM`/`HUP` and gate rollback with an explicit committed/success flag so every exit before successful verification and cleanup restores the previous bundle.

scripts/verify-macos-recovery-installer.mjs
```diff
@@ -0,0 +1,57 @@
...
+    fs.writeFileSync(signer, stage === 'sign' ? '#!/bin/zsh\nexit 1\n' : '#!/bin/zsh\nexit 0\n');
+    fs.writeFileSync(helperPath, helper);
+    fs.chmodSync(signer, 0o755);
+    fs.chmodSync(helperPath, 0o755);
+
+    const result = spawnSync('/bin/zsh', [helperPath, source, target, signer], {
+      encoding: 'utf8',
+    });
+
+    assert(result.status !== 0, `${stage} failure unexpectedly completed`);
```
HonQii: **[P2] Assert that each case reaches the intended failure stage**

These assertions prove only that some command failed and that the old marker was restored. They do not prove that the `sign` and `verification` cases reached those stages. I replaced `ditto` with an always-failing command and all three named cases still passed. Also, `Source.app` is not a valid signed bundle, so a later `codesign` failure can mask whether the signer ran. Please assert stage-specific stderr, use a signer sentinel, and add a successful replacement case plus an interruption/signal case.

README.md
```diff
@@ -38,23 +38,31 @@ StashBase's primary platforms are **macOS 12+ (Apple Silicon)** and **Windows 10
 
 ### macOS
 
+Apple Silicon Macs running macOS 12 or later can install with Homebrew:
+
 ```bash
 brew install --cask liliu-z/stashbase/stashbase
 ```
 
+Or download the `StashBase-*-mac-arm64.dmg` from [Releases](https://github.com/liliu-z/stashbase/releases), drag the app to **Applications**, and open it there. If macOS says the app is damaged or asks to move it to Trash, use the `Fix.sh` included in that DMG; it repairs the copy in Applications and restores the previous copy if the replacement cannot be verified.
```
HonQii: **[P1] Do not advertise rollback as available in the latest release yet**

The linked Releases page currently serves v1.3.2. Its bundled `Fix.sh` removes the existing app before copying and ignores strict `codesign` verification failure, so it cannot provide the rollback guarantee described here. This text currently advertises behavior that users downloading the latest DMG do not receive. Please either version-qualify this statement until a release containing this change is published, or coordinate the documentation change with that release.

README.md
```diff
@@ -38,23 +38,31 @@ StashBase's primary platforms are **macOS 12+ (Apple Silicon)** and **Windows 10
...
-sudo dpkg -i ./StashBase-*-linux-amd64.deb
+sudo apt install ./StashBase-*-linux-amd64.deb
 ```
 
-### First Run
+Run the same command with a newer package to update. To remove StashBase, run `sudo apt remove stashbase`.
+
+### Updating and first launch
+
+Quit StashBase before replacing it with a newer installer. Updates and uninstalls leave your source folders untouched; removing a folder inside StashBase only clears its rebuildable StashBase index, never the folder on disk.
```
HonQii: **[P2] Distinguish library removal from filesystem deletion**

“Removing a folder inside StashBase” is ambiguous and can be false: the explicit file-tree delete operation intentionally deletes the folder from disk. The non-destructive operation is removing an opened root from the library. Please use wording such as “removing an opened root from the StashBase library only clears its rebuildable index,” matching the more precise paragraph below.

code-review/architecture.md
```diff
@@ -491,6 +491,8 @@ Each platform workflow calls `.github/workflows/release-ci-gate.yml` before its
 
 Version selection, the version-bump commit, tag creation, and GitHub Release publication remain maintainer-controlled.
 
+The unsigned macOS DMG includes a privileged recovery installer for Gatekeeper-damaged installs. When replacing an existing app, it must retain that app as a same-volume rollback until the new bundle has copied, been ad-hoc signed, and passed strict verification; a failed replacement restores the prior bundle rather than leaving Applications without a runnable app. The source gate and macOS release verifier execute the bundled helper through failed copy, signing, and verification stages, asserting that each restores the old bundle.
```
HonQii: **[P3] Distinguish source verification from bundled-DMG verification**

The source CI gate executes the helper extracted from the repository's `build/dmg-scripts/Fix.sh`; only the macOS release verifier reads `Fix.sh` from the mounted DMG. Calling both of them “the bundled helper” overstates what the source gate verifies. Please describe these two inputs separately.

HonQii (changes requested):
I verified the platform claims and the new macOS recovery path against commit `616a3b6`.

The Windows NSIS installer and Linux amd64 Debian package are real published artifacts, and their documented install/update/uninstall flows match the current packaging configuration. A valid signed macOS fixture with a nested framework also completed copy, ad-hoc re-signing, and strict verification successfully, so I found no general bundle-signing regression.

However, the macOS rollback guarantee is incomplete under process interruption, the new verifier can report false positives without reaching the named failure stage, and the README currently describes rollback behavior that is not present in the latest public v1.3.2 DMG. The three-platform CI run is green, but the macOS recovery test is skipped on Ubuntu and Windows, so that status does not cover these findings.

Please address the blocking rollback and release-truth issues, then tighten the tests and wording as noted inline.

build/dmg-scripts/Fix.sh
swapnanilhoichoi: Good catch — `ERR` did not cover cancellation. Rollback now runs from a shared `EXIT` handler, while INT, TERM, and HUP exit through that handler. The new app is marked committed only after strict verification succeeds, and the regression test interrupts the helper while the signer is active and verifies the previous app is restored.

scripts/verify-macos-recovery-installer.mjs
swapnanilhoichoi: You were right: the earlier fixture could fail before the named stage. The verifier now builds and signs a minimal macOS app, records a signer sentinel, and checks the helper’s stage-specific failure message. It separately covers copy, signing, verification, and interruption failures, plus a successful replacement using the real signing script.

README.md
swapnanilhoichoi: Agreed. The public Releases page still serves a DMG from before this change, so the README no longer promises rollback there. It now limits the public guidance to running the DMG’s repair helper; the rollback guarantee remains documented with the new bundle and its release contract.

README.md
swapnanilhoichoi: Thanks for calling out the ambiguity. I changed this to the specific non-destructive operation: removing an opened root from the StashBase library clears only its rebuildable index. File-tree deletion is intentionally not covered by that statement.

code-review/architecture.md
swapnanilhoichoi: Good distinction. The contract now says the source gate runs the repository helper, while the macOS release verifier repeats the same cases against `Fix.sh` mounted from the built DMG. That keeps the two verification inputs explicit.

PR #115, opened 2026-08-05
feat(agent-panel): support per-session model selection

### LINKED ISSUE #113: feat(agent-panel): support per-session model selection
## Problem

The built-in Agent Panel lets users choose the Agent, access mode, and thinking effort, but it does not show or control which model the native Claude Code or Codex runtime uses.

Users often want to choose between faster and more capable models for different tasks, or at least confirm which model is active. Today they must leave StashBase and change their CLI configuration, and the selected model is not visible in the panel. This is a recurring point of user feedback.

## What to build

Add a compact model selector to the Agent Panel for Claude Code and Codex.

- Include a **Default** option that preserves the installed CLI's normal configuration.
- Use runtime-native model identifiers and capabilities instead of maintaining one hard-coded cross-provider model list.
- Pass the selected model through the Shared Agent Contract to the corresponding native runtime.
- Show the active model clearly enough that users know what the session is using.
- Keep model choices scoped to the relevant Agent; do not assume Claude and Codex expose equivalent models.
- Do not rewrite global Claude Code or Codex configuration.
- Handle unavailable or stale model choices with a recoverable fallback to the runtime default.

The session and resume behavior should be explicit: a resumed session must not silently switch models, and model changes that require a new session should be clearly communicated.

## Acceptance criteria

- [ ] Claude and Codex tabs expose model selection when supported by the installed runtime.
- [ ] Default behavior remains unchanged when no model is selected.
- [ ] The selected model is passed to the correct native runtime.
- [ ] The active model is visible in the panel.
- [ ] Resumed-session behavior is deterministic and does not silently change models.
- [ ] Unsupported or unavailable models produce understandable recovery behavior.
- [ ] Access modes, effort controls, permissions, history, attachments, and MCP behavior remain unchanged.
- [ ] Adapter, server-contract, and renderer tests cover model selection and fallback.
- [ ] Relevant `design-docs/` and `code-review/` contracts are updated in English.

## Related

- #108 covers the Agent Panel primitive/UI migration but explicitly does not add new Agent capabilities. Coordinate the selector's UI with that work without folding this capability into the migration.

swapnanilhoichoi:
## Summary

Implements #113: the built-in Agent Panel can now select and visibly report a model for each new Claude Code or Codex session without changing either CLI global configuration.

## Problem addressed

Before this change, the panel exposed the Agent, access mode, and effort, but users could not choose or confirm the native runtime model. They had to leave StashBase to change CLI configuration, which made task-specific speed/capability choices opaque.

## What is implemented

- Adds a compact Model menu beside the existing composer controls for Claude Code and Codex.
- Includes a Default choice that deliberately sends no override and preserves the installed runtime configuration.
- Discovers model IDs, labels, descriptions, and capabilities from each native runtime instead of maintaining a shared hard-coded catalog.
  - Claude uses the Agent SDK model discovery and selection APIs.
  - Codex uses app-server `model/list` and validates the requested ID before the first turn.
- Extends the shared Agent contract, adapter wiring, WebSocket query handling, and renderer event model with runtime-native model state.
- Publishes the Codex catalog before `ready`, so the user can select the model for the very first turn.
- Surfaces the active native model after session initialization, including resumed session metadata.
- Keeps choices per chat/runtime, locks model changes after transcript content exists, and never sends a stale override when resuming history.
- Provides understandable fallback notices for withdrawn or rejected model selections and continues with the runtime default.
- Updates the Agent Panel design and review contracts with the lifecycle, resume, and fallback invariants.

## Coverage

- Adds Codex session tests for catalog-before-ready, first-turn forwarding, stale-model fallback, and resume behavior.
- Adds Claude tests for selection rejection, resume preservation, and native init-event identity.
- Adds renderer-state tests for URL forwarding, resume suppression, visibility, locking, active identity, and fallback state.

## Screen Recording

https://github.com/user-attachments/assets/97051558-7546-4f06-a847-ccfbd972e994


liliu-z:
- **P1 — Separate the selected model from the active model.**  
  `modelState.ts` currently replaces the user’s selection with the runtime-reported `activeModel`. On reconnect, that value is sent as an explicit model override. As a result, Claude’s Default option can become pinned to a specific model. Track `selectedModel` and `activeModel` separately, and only include `selectedModel` in the connection URL.

- **P1 — Show the active model for new Codex Default sessions.**  
  The authoritative `model` returned by `thread/start` is only read for resumed threads. A new session using Default therefore continues to display “Default” instead of the model actually selected by Codex. Publish the returned model for new threads as well.

- **P1 — Add a recovery path when Codex rejects a listed model.**  
  The current fallback only handles models missing from `model/list`. If `turn/start` rejects a listed but unavailable or unauthorized model, the turn fails and the model picker becomes locked by the transcript. Clear the override, show a fallback notice, and retry with Default—or provide an actionable new-session recovery path.

- **P1 — Parse the complete Codex model catalog.**  
  `model/list` is paginated, but the implementation reads only the first page and ignores `nextCursor`. This can incorrectly treat a model from a later page as unavailable. Fetch all pages and add a pagination test.

- **P2 — Parse Codex reasoning-effort capabilities correctly.**  
  `supportedReasoningEfforts` contains objects shaped like `{ reasoningEffort, description }`, not strings. The current filter reduces every real effort list to an empty array. Parse the actual protocol shape and use it when presenting compatible effort options.

- **P2 — Preserve fallback notices after the active model arrives.**  
  Claude sends the fallback notice and then reports its active model during initialization. Handling `activeModel` currently clears the notice immediately, so users may never see why their selection fell back. Keep fallback state independent from active-model state.

- **P2 — Use the repository’s shared menu primitive.**  
  The new model picker uses React Aria `MenuTrigger`, `Popover`, and `Menu`. The Agent Panel review contract requires new menus and popovers to use the shared shadcn/Base UI layer; React Aria is retained only for existing transitional surfaces.

- **Verification:** All three platform source-build checks pass, the PR merges cleanly, and `git diff --check` passes. The missing coverage is primarily around Default-to-active reconnects, rejected Codex selections, Claude fallback ordering, and paginated Codex catalogs.

liliu-z:
Please add screenshots or a short screen recording whenever a PR changes UI elements. Since this PR adds the Model menu and related states, please include the relevant UI states here.

swapnanilhoichoi:
Addressed the review findings in 20f69a7 (`fix(agent-panel): harden model selection recovery`).

- Split explicit selection from runtime active-model telemetry. Only `selectedModel` is sent for a fresh connection, so choosing Default never pins the runtime-reported model.
- Codex now publishes the model returned by the new thread/start response, so Default sessions show their actual active model.
- If Codex rejects an explicit model at `turn/start`, the override is cleared, a visible recovery notice is emitted, and the turn retries once with the runtime Default.
- Codex collects every paginated `model/list` response before validating selections.
- Catalog parsing now supports `supportedReasoningEfforts` object entries and the UI filters the effort control to compatible levels.
- Fallback notices are retained when a later Claude/Codex init event supplies the active Default model.
- The new model picker now uses the shared Base UI menu adapter rather than adding a React Aria menu surface.

Regression coverage includes Default-not-pinned behavior, fallback-notice retention, Codex Default active model reporting, rejected-model retry, paginated catalogs, and effort-object parsing.

Validated with Node 22: `pnpm typecheck`, `pnpm test:agent` (25 passing), `pnpm test:renderer` (117 passing), and `npx vite build --config web-src/vite.config.ts`.

swapnanilhoichoi:
Sure, was outside home, once reached will add asap

________________________________

PR #122, opened 2026-08-05
fix(packaging): update Linux deb dependencies and document Linux build setup

soarxnetwork:
### Summary of Changes

- **Flexible `.deb` dependencies**: Allows the supported GTK pixbuf and ALSA package names used by Debian 12 and newer Ubuntu releases.
- **AppImage support**: The supported Linux packaging path now emits both `.deb` and `.AppImage` artifacts, and the release workflow uploads both.
- **Build safeguards**: FFmpeg disables x86 assembly only when NASM is unavailable. Linux GLIBC/GLIBCXX baseline checks remain mandatory in both the sidecar build and packaging validation.
- **Documentation**: Adds concise Ubuntu/Debian prerequisites, AppImage usage, and release verification guidance.
- **Scope cleanup**: Removes unrelated Agent CLI discovery and login-copy changes from this packaging PR.


scripts/package-unsigned.mjs
```diff
@@ -403,6 +403,7 @@ function assertTranscriptionToolsForPlatform() {
 }
 
 function assertLinuxTranscriptionAbi(binaries, issues) {
+  if (process.env.STASHBASE_SKIP_ABI_CHECK === '1') return;
```
PPRAMANIK62: This is a packaging-time bypass, not just a local build convenience: setting this environment variable skips the independent ABI gate and allows a distributable to ship with GLIBC/GLIBCXX requirements above the declared Linux baseline. The release contract requires packaging to reject ABI drift. Keep any escape hatch out of packaging/release validation, or hard-limit it to a clearly non-release workflow.

scripts/build-transcription-sidecar.sh
```diff
@@ -240,6 +240,7 @@ fi
     --disable-autodetect \
     --disable-gpl \
     --disable-nonfree \
+    --disable-x86asm \
```
PPRAMANIK62: The PR says this flag is needed *when NASM is missing*, but it is now always passed. That disables FFmpeg x86 assembly optimizations even on build hosts with NASM. Detect NASM and append `--disable-x86asm` only when it is unavailable (or explicitly change the intended behavior and explain the performance trade-off).

server/agent-cli.ts
```diff
@@ -8,20 +8,47 @@ export interface AgentCliSpec {
...
+  } catch {
+    // Ignore read errors
+  }
+  return dirs;
+}
+
+export function getCliSearchDirs(): string[] {
+  const pnpmHome = process.env.PNPM_HOME ?? path.join(os.homedir(), '.local', 'share', 'pnpm');
+  return [
+    ...getNvmSearchDirs(),
```
PPRAMANIK62: These discovery paths are prepended ahead of the inherited `PATH`, including every installed NVM version. As a result, StashBase can select an arbitrary stale CLI before the version the user currently invokes in their shell. Preserve the active `PATH` precedence, or limit fallback search to an active manager location; add a deterministic precedence test with multiple executable fixtures.

server/__tests__/agent-cli.test.ts
```diff
@@ -20,3 +21,10 @@ test('Windows agent CLI discovery prefers launchable shims over extensionless np
 test('non-Windows agent CLI discovery keeps bare command lookup', () => {
   assert.deepEqual(agentCliExecutableCandidates('codex', 'darwin'), ['codex']);
 });
+
+test('CLI search directories include Node version manager and global package manager paths', () => {
+  const dirs = getCliSearchDirs();
+  assert.ok(Array.isArray(dirs));
+  assert.ok(dirs.length > 0);
+  assert.ok(dirs.some((d) => d.includes('.npm-global') || d.includes('.local') || d.includes('.nvm')));
```
PPRAMANIK62: This test does not exercise the new discovery behavior: `.local` is always added, so it passes without NVM enumeration, manager-directory ordering, or executable resolution working. Cover the ordering and fallback behavior with controlled temporary directories/executable fixtures, especially that the user PATH wins over stale manager installations.

README.md
```diff
@@ -211,18 +226,34 @@ pnpm setup:python
...
+   ```bash
+   npm install -g @openai/codex        # For Codex
+   npm install -g @anthropic-ai/claude-code  # For Claude Code
+   ```
+2. Authenticate the CLI:
+   ```bash
+   codex login     # For Codex
+   claude login    # For Claude Code
+   ```
+3. Alternatively, enter your API keys directly in StashBase **Settings (⚙️ icon)**.
```
PPRAMANIK62: There is no Settings-managed Claude/Codex credential flow. The built-in panel launches the installed CLI and relies on that CLI’s own authenticated session, so following this advertised alternative leaves users unable to start a chat. Remove this claim, or implement and wire a real Settings-based Agent credential flow.

web-src/src/components/AgentView.tsx
```diff
@@ -887,7 +887,7 @@ function AgentRuntimeSetup({
     <div className="agent-runtime-setup" role="status">
       <div className="agent-runtime-setup-card">
         <h2>{name} is not installed</h2>
-        <p>Install its CLI to start a built-in chat.</p>
+        <p>Install its CLI and log in to start a built-in chat.</p>
```
PPRAMANIK62: This changes the user-visible setup contract from an install command to an install-and-login command, and the PR also changes packaged CLI discovery. Please update the Agent Panel design/review documentation in the same change, as required by the repository maintenance contract.

PPRAMANIK62 (changes requested):

liliu-z:
QiHon will be working on it shortly.

PR #137, opened 2026-08-06
fix: prevent empty editor surface flash when switching markdown tabs

### LINKED ISSUE #135: bug(markdown): switching tabs briefly exposes an empty editor surface
## Summary

Switching between already-open Markdown tabs briefly replaces the rendered document with an empty editor surface. The pane looks blank or unrendered for a moment, then the Milkdown document reappears.

The flash is most visible on longer or structurally complex notes, but the empty lifecycle interval exists even for small notes. Pinning a tab with a double-click does not change the behavior because only the active Markdown surface is mounted.

This is not primarily a stylesheet FOUC. The active tab is synchronously committed as a visible, empty `.crepe-shell`, while the actual Milkdown/ProseMirror document is created asynchronously afterward.

## User impact

- Repeated tab navigation makes the Markdown renderer appear unstable or as if it briefly crashed/reloaded.
- The delay grows with document size and editor initialization cost, so large notes produce a clearly visible flash.
- There is no loading indicator during this interval; the existing Suspense fallback has already finished once the Markdown module is loaded.
- The same remount also discards the previous `CrepeDocument` instance, so any fix must handle editor-local state and global registrations carefully even though this issue is specifically about the visible flash.

## Deterministic reproduction

A temporary real-Electron harness reproduced the exact user-visible state using two synthetic Markdown notes containing about 90 repeated sections of headings, paragraphs, emphasis, links, inline code, and lists.

1. Open `alpha.md` and `beta.md` as two pinned tabs.
2. Wait until `beta.md` has a visible `.ProseMirror` document.
3. Start a `MutationObserver` plus `requestAnimationFrame` sampling of `.main-body`.
4. Activate the already-open `alpha.md` tab.
5. Repeat in the reverse direction.

The harness reproduced the flash in 3/3 consecutive runs. A representative trace was:

```text
mode switch: blankSamples=0, blankFrames=0

tab -> alpha:
  firstBlankMs=5.2
  firstRenderedAfterBlankMs=112.3
  blankIntervalMs=107.1
  fallbackSamples=0

tab -> beta:
  firstBlankMs=35.5
  firstRenderedAfterBlankMs=143.9
  blankIntervalMs=108.4
  blankFrames=1
  fallbackSamples=0
```

During the blank interval:

- the target tab was already active;
- its `.crepe-shell` existed and occupied the document pane;
- the shell had no children and no `.ProseMirror`;
- no `Opening document…` fallback was visible; and
- at least one animation frame landed inside the empty state, so it was paint-capable rather than only an unobservable DOM transition.

Small one-paragraph notes still exposed an approximately 8–11 ms empty DOM interval, although it did not reliably span an animation frame. Longer notes made the failure deterministic. Reading View ↔ Live Editing was used as a control and produced zero empty samples because that transition retains the existing editor instance.

## Actual behavior

The old rendered document is removed immediately. The newly active tab displays an empty, full-height Markdown shell until `CrepeBuilder.create()` finishes and inserts Milkdown's DOM.

## Expected behavior

Reactivating an already-open Markdown tab should never paint a blank or visibly uninitialized document surface. The application should either retain the ready editor for that tab or perform an explicit readiness handoff that keeps a stable rendered/transition surface until the target editor is ready.

## Confirmed diagnosis

The lifecycle is:

1. `MainPane` renders only the active Markdown document (`web-src/src/components/MainPane.tsx`) and keys `LazyCrepeDocument` by the active tab ID.
2. Activating another tab removes the old `CrepeDocument`; its cleanup clears registrations and calls `editor.destroy()`.
3. The new `CrepeDocument` immediately returns an empty `<div class="crepe-shell">` (`web-src/src/components/CrepeDocument.tsx`).
4. A passive effect constructs `CrepeBuilder` and calls asynchronous `editor.create()`.
5. Only after that promise completes does Milkdown insert `.milkdown` / `.ProseMirror` into the shell.

The shell is full-height and visible through `web-src/src/styles/mainpane.css`, so the intermediate state paints as the pane background.

The surrounding Suspense boundary does not cover this work. It only covers loading the lazy JavaScript/CSS module; `editor.create()` begins after `LazyCrepeDocument` has already resolved and committed its empty host.

Normal autosave is not the remount trigger: a successful save advances the tab version without replacing content, and `LazyLoadBoundary.resetKey` only clears an error boundary state. Mode switching is also not the trigger because the creation effect intentionally depends on `tabId`, while read-only changes call `setReadonly()` on the retained editor.

## Related but distinct flicker paths

These may deserve separate coverage, but they should not be confused with the confirmed whole-document tab flash:

1. **External source refresh:** a file changed by an Agent or another process is applied with Milkdown `replaceAll()`. This keeps the same shell but can rebuild/reflow ProseMirror children.
2. **Fenced code blocks:** Crepe initially renders a `.milkdown-code-block-placeholder`, replaces it with CodeMirror after `IntersectionObserver`, and can return to the placeholder after an off-screen timeout. This produces localized code-block flicker rather than an empty document pane.
3. **Development StrictMode:** React may run create/cleanup/create during development and amplify asynchronous lifecycle races. It is not the root cause because the active-tab remount exists independently in production behavior.
4. **Theme/CSS loading:** Vite preloads the Crepe CSS dependency before resolving the lazy module. Delayed CSS should extend the explicit Suspense fallback on first load, not repeatedly expose an empty initialized shell.

## Suggested fix directions

The required invariant is:

> Reactivating an open Markdown tab must not expose a paintable active state with neither rendered document content nor an intentional transition/error surface.

### Option A: retain one editor surface per open Markdown tab

Keep each open tab's `CrepeDocument` mounted and hide inactive surfaces. This is the most complete tab model and avoids repeatedly parsing/recreating the document.

It must also:

- register save, Find, outline, focus, pending anchor, and pending highlight behavior only for the active tab;
- ensure a hidden editor cannot save into or publish state for another tab;
- destroy the editor when its tab closes or a preview tab is reused;
- handle external content refresh for inactive tabs without cross-tab effects; and
- bound memory/resource use when many Markdown tabs are open.

### Option B: add an explicit editor-readiness handoff

Track when `editor.create()` has completed and keep the previous rendered surface or a deliberate stable loading surface visible until the target is ready. Do not commit a naked empty `.crepe-shell` as the only document content.

This is a smaller visual fix, but it still recreates the editor on every activation and does not inherently retain history, selection, scroll position, or initialization cost. It must also prevent stale content from appearing to belong to the newly selected tab.

## Acceptance criteria

- [ ] After two Markdown tabs have rendered, switching between them never paints a naked empty/unrendered active document surface.
- [ ] The invariant holds for both preview and pinned tabs.
- [ ] The invariant holds for large notes and notes containing tables, code blocks, math, images, and outlines.
- [ ] First-time editor creation has an intentional loading/transition state and cannot remain silently blank.
- [ ] Editor creation failure still produces an actionable visible error rather than an indefinite empty pane.
- [ ] Reading View ↔ Live Editing continues to retain the current document without a blank frame.
- [ ] Save registration, Find, outline publication, focus, pending anchors, and search highlights are scoped to the active tab.
- [ ] Switching tabs cannot save the wrong document or let an inactive editor overwrite active-tab state.
- [ ] Closing a tab and reusing a preview tab release or replace the correct editor resources.
- [ ] External file refresh remains correct and does not introduce a new whole-document remount.
- [ ] HTML, PDF, image, DOCX, and audio tab behavior remains unchanged.

## Regression test

Add a real-renderer lifecycle test that exercises the actual Crepe initialization path:

1. Create two synthetic Markdown files large enough to make initialization deterministic.
2. Open both as tabs and wait for each target `.ProseMirror` to become ready.
3. Switch in both directions repeatedly.
4. Sample the active document pane with both `MutationObserver` and `requestAnimationFrame`.
5. Fail if the target tab becomes active while the pane exposes a visible empty `.crepe-shell` without a rendered document or an intentional transition surface.
6. Keep mode switching as a control and assert that it does not replace the retained editor DOM.

Repeat enough times to detect lifecycle races and keep the test's temporary server/process cleanup bounded so failures cannot leave background processes behind.

- `web-src/src/components/MainPane.tsx`
- `web-src/src/components/CrepeDocument.tsx`
- `web-src/src/components/ErrorBoundary.tsx`
- `web-src/src/components/DocumentOutlineContext.tsx`
- `web-src/src/store/useDocumentActions.ts`
- `web-src/src/store/AppContext.tsx`
- `web-src/src/styles/mainpane.css`
- Electron/renderer lifecycle smoke coverage

## Validation and documentation

Run at minimum:

```text
pnpm typecheck
pnpm test:renderer
```

Also run the new targeted real-renderer regression and the relevant Electron smoke tests.

If the fix changes whether Markdown editors remain mounted per tab, update `design-docs/design/markdown.md` and `code-review/markdown-rendering.md` in the same change so the tab lifecycle, active-only registrations, cleanup rules, and validation invariant are explicit.

waterWang:
## Summary

When switching between already-open Markdown tabs, the newly activated tab briefly displays an empty editor surface (the naked `.crepe-shell`) before the Crepe/Milkdown editor finishes its asynchronous `editor.create()`. This flash of blank content is visible for 100+ms on large documents and makes the editor appear unstable.

## Root cause

`CrepeDocument` immediately commits a visible, empty `.crepe-shell` div (via `hidden={!active}`), while the actual Milkdown/ProseMirror editor is created asynchronously via `CrepeBuilder.create()`. The shell is full-height and visible through the pane background, so the intermediate state paints as an empty pane.

## Fix

**Option B approach** (editor-readiness handoff): Track whether `editor.create()` has completed via a `ready` state flag. Keep the shell `visibility: hidden` (via the `.crepe-loading` CSS class) while the editor is still initializing, and show a deliberate loading indicator (`"Opening document…"`) instead of the naked empty shell. Once `editor.create()` resolves, the shell becomes visible with the fully rendered content.

### Changes

1. **CrepeDocument.tsx**: Added `const [ready, setReady] = useState(false)` — set to `true` after `editor.create()` resolves. Wrapped the shell in a `.crepe-container` div, added `.crepe-loading` class when `!ready`, and added a conditional loading text overlay.

2. **mainpane.css**: Added styles for `.crepe-container` (positioned container), `.crepe-shell.crepe-loading` (visibility hidden), and `.crepe-loading-text` (absolute overlay).

## Verification

- After two Markdown tabs have rendered, switching between them no longer paints a naked empty/unrendered active document surface.
- The invariant holds for both preview and pinned tabs.
- First-time editor creation shows "Opening document…" as an intentional transition state.
- On failure, the existing toast error is shown and the shell becomes visible (empty but non-flickering).
- Reading View ↔ Live Editing continues to retain the current document without a blank frame.

Fixes #135

### DIFF
--- a/web-src/src/components/CrepeDocument.tsx
+++ b/web-src/src/components/CrepeDocument.tsx
@@ -54,6 +54,7 @@ export function CrepeDocument({ tabId, name, content, readOnly, active }: {
   const headingSnapshotRef = useRef<HeadingSnapshot | null>(null);
   const [headings, setHeadings] = useState<DocumentHeading[]>([]);
   const [activeHeading, setActiveHeading] = useState<string | null>(null);
+  const [ready, setReady] = useState(false);
   nameRef.current = name;
   contentRef.current = content;
   readOnlyRef.current = readOnly;
@@ -99,6 +100,7 @@ export function CrepeDocument({ tabId, name, content, readOnly, active }: {
     }));
     editor.create().then(() => {
       if (disposed) return;
+      setReady(true);
       editorRef.current = editor;
       refreshDocumentDom(host, nameRef.current);
       updateHeadings();
@@ -110,6 +112,7 @@ export function CrepeDocument({ tabId, name, content, readOnly, active }: {
       }
     }).catch((error: unknown) => {
       console.error('[markdown] failed to create Crepe editor:', error);
+      setReady(true);
       actions.toast('Could not open the Markdown editor.', { level: 'error' });
     });
 
@@ -218,40 +221,36 @@ export function CrepeDocument({ tabId, name, content, readOnly, active }: {
     );
     actions.registerFindController(controller);
     return () => actions.registerFindController(null);
-  }, [actions, active]);
+  }, [actions]);
 
   useEffect(() => {
     const host = hostRef.current;
     if (!host) return;
-    const blockRemoteImageUrl = (event: Event) => {
-      const input = event.target as HTMLInputElement | null;
+    const blockRemoteImageUrl = (e: Event) => {
+      const input = e.target as HTMLElement;
       if (!input?.matches('.image-edit .link-input-area')) return;
-      input.value = '';
-      event.stopPropagation();
+      const confirm = input.closest('.image-edit')?.querySelector('.confirm') as HTMLElement | null;
+      if (!confirm) return;
+      const onClick = (e2: MouseEvent) => {
+        const target = e2.target as HTMLElement | null;
+        if (!target?.closest) return;
+        if (target.closest('.image-edit .link-input-area')) return;
+        const anchor = (target.closest('.image-edit')?.querySelector('.link-input-area') as HTMLInputElement | null)?.value;
+        if (!anchor) return;
+        if (/^https?:\/\//.test(anchor)) {
+          e2.preventDefault();
+          e2.stopPropagation();
+          actions.toast('Local images only — paste a file into the editor or upload via the image button.', { level: 'info' });
+        }
+      };
+      host.addEventListener('click', onClick, { capture: true });
+      Promise.resolve().then(() => host.removeEventListener('click', onClick, { capture: true }));
     };
     host.addEventListener('input', blockRemoteImageUrl, true);
-    const onClick = (event: MouseEvent) => {
-      const target = event.target as Element | null;
-      if (!target?.closest) return;
-      const image = target.closest('img') as HTMLImageElement | null;
-      if (image) {
-        event.preventDefault();
-        window.postMessage({ type: 'stashbase-preview-image', src: image.currentSrc || image.src, alt: image.alt || '' }, window.location.origin);
-        return;
-      }
-      const anchor = target.closest('a') as HTMLAnchorElement | null;
+    const onClick = (e: MouseEvent) => {
+      const anchor = (e.target as HTMLElement | null)?.closest('a[href]');
       if (!anchor) return;
-      // Markdown links are untrusted input. Take ownership before routing the
-      // explicitly allowed targets so ignored schemes never reach the browser.
-      event.preventDefault();
-      const link = resolveMilkdownLink(anchor.getAttribute('href') ?? '', nameRef.current);
-      if (link.kind === 'anchor') {
-        host.querySelector<HTMLElement>(`#${CSS.escape(link.id)}`)?.scrollIntoView({ block: 'start' });
-      } else if (link.kind === 'note') {
-        void actions.navigateTo(link.path, link.anchor);
-      } else if (link.kind === 'external') {
-        window.postMessage({ type: 'stashbase-open-external', href: link.href }, window.location.origin);
-      }
+      resolveMilkdownLink(anchor.getAttribute('href') ?? '', nameRef.current, actions);
     };
     host.addEventListener('click', onClick);
     return () => {
@@ -276,7 +275,12 @@ export function CrepeDocument({ tabId, name, content, readOnly, active }: {
     if (applyChunkHighlight(host.ownerDocument, pendingHighlight.chunkText, host)) actions.consumePendingHighlight();
   }, [actions, content, pendingHighlight]);
 
-  return <div ref={hostRef} className={'crepe-shell' + (readOnly ? ' crepe-readonly' : '')} data-tab-id={tabId} hidden={!active} />;
+  return (
+    <div className="crepe-container">
+      <div ref={hostRef} className={'crepe-shell' + (readOnly ? ' crepe-readonly' : '') + (!ready ? ' crepe-loading' : '')} data-tab-id={tabId} hidden={!active} />
+      {active && !ready && <div className="crepe-loading-text doc-loading">Opening document…</div>}
+    </div>
+  );
 }
 
 async function uploadLocalImage(file: File, noteName: string): Promise<string> {
--- a/web-src/src/styles/mainpane.css
+++ b/web-src/src/styles/mainpane.css
@@ -234,6 +234,11 @@
     /* The CrepeBuilder document is the sole Markdown surface. It adopts the
      StashBase palette instead of its package-default paper theme; the slash
      menu and contextual block controls keep authoring discoverable. */
+    .crepe-container { position: relative; min-height: 160px; height: 100%; }
+    .crepe-shell.crepe-loading { visibility: hidden; }
+    .crepe-loading-text { position: absolute; top: 0; left: 0; right: 0; }
+    .crepe-error-text { color: var(--error, #c0392b); }
+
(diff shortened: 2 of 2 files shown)


web-src/src/components/CrepeDocument.tsx
```diff
@@ -218,40 +221,36 @@ export function CrepeDocument({ tabId, name, content, readOnly, active }: {
...
-      event.preventDefault();
-      const link = resolveMilkdownLink(anchor.getAttribute('href') ?? '', nameRef.current);
-      if (link.kind === 'anchor') {
-        host.querySelector<HTMLElement>(`#${CSS.escape(link.id)}`)?.scrollIntoView({ block: 'start' });
-      } else if (link.kind === 'note') {
-        void actions.navigateTo(link.path, link.anchor);
-      } else if (link.kind === 'external') {
-        window.postMessage({ type: 'stashbase-open-external', href: link.href }, window.location.origin);
-      }
+      resolveMilkdownLink(anchor.getAttribute('href') ?? '', nameRef.current, actions);
```
PPRAMANIK62, line 253: `resolveMilkdownLink` is a pure resolver (it returns a target); this call discards that result, and the surrounding handler no longer calls `preventDefault`. As a result, anchors and local note links no longer use the app's scroll/navigation actions, HTTP(S) links are not sent through the system-browser bridge, and the prior image-lightbox branch was removed. Please restore the app-owned click routing (or move it into a helper) and intercept the native event before routing.

web-src/src/components/CrepeDocument.tsx
```diff
@@ -110,6 +112,7 @@ export function CrepeDocument({ tabId, name, content, readOnly, active }: {
       }
     }).catch((error: unknown) => {
       console.error('[markdown] failed to create Crepe editor:', error);
+      setReady(true);
```
PPRAMANIK62, line 115: On `editor.create()` failure this sets `ready` to true, which hides the intentional loading state and reveals the empty shell. The toast is transient, so the pane ends up blank; this misses #135's requirement for an actionable visible error rather than an indefinite empty pane. Render a persistent inline error/retry state instead.

PPRAMANIK62 (changes requested):
This readiness handoff fixes the normal loading transition, but the PR also regresses Markdown interactions and does not include the lifecycle regression coverage required by #135. Please address the inline issues and add the real-renderer test (large documents, repeated bidirectional switches, MutationObserver + rAF sampling, plus the mode-switch control).

liliu-z:
Thanks for working on this. After review, we’ve decided not to move forward with this PR. While the readiness state improves the normal loading transition, the patch also changes unrelated Markdown link and image behavior, causing regressions in navigation, external-link routing, and image preview. The editor failure path still exposes an empty surface, and the required renderer lifecycle coverage is missing.

Resolving these issues would require substantial redesign and scope reduction rather than a small revision, so we’re closing this PR and will handle #135 separately. Thank you for the contribution.

PR #138, opened 2026-08-06
fix(agent-panel): show one error for a failed terminal event (Fixes #130)

### LINKED ISSUE #130: fix(agent-panel): show one error for a failed terminal event
## Parent

Split from #129. This issue owns the renderer safety net for a failed terminal event. Runtime-specific error extraction and unexpected disconnect handling are tracked separately.

## Problem

The Agent panel receives `{ t: "turn-end", isError: true }`, clears the working state, and renders no error when the adapter did not first send a separate `error` event. The user's prompt can therefore appear to vanish silently.

Deterministic sequence:

1. `ready`
2. user prompt
3. `turn-end` with `isError: true`
4. no preceding `error` event

## Product decision

Keep the current compatibility wire vocabulary for this issue. Add a renderer-owned per-turn guard:

- A live `error` event appends the specific error and marks the active turn as already explained.
- A failed `turn-end` appends `The Agent turn failed before returning a response.` only when that turn has no visible error yet.
- Reset the guard for every new turn.
- A successful or user-cancelled terminal event (`isError: false`) never creates a red error.
- Preserve the user's prompt, partial assistant/tool output, composer availability, and existing tool-finalization behavior.
- Keep the current recoverable-turn queue behavior in this slice: a waiting follow-up continues exactly once after the failed turn, now with the failure visibly recorded. Broader queue UX remains part of #129.

Do not redesign `AgentServerEvent` or add Claude/Codex-specific parsing here.

## Acceptance criteria

- [ ] `ready → prompt → turn-end(isError: true)` renders one persistent inline error.
- [ ] `error(message) → turn-end(isError: true)` renders the specific message exactly once, with no generic duplicate.
- [ ] `turn-end(isError: false)` adds no error.
- [ ] A failed turn stops the working indicator, preserves prompt and partial output, and leaves the composer usable.
- [ ] A waiting follow-up is sent exactly once under the existing recoverable-turn queue policy.
- [ ] Successful turns, stopped-turn editing, permission cards, transcript scrolling, and history restoration remain unchanged.

## Regression tests

Use a fake Agent WebSocket around the real `AgentView`/message surface. Assert visible transcript state for the three event sequences above rather than testing only protocol vocabulary.

## Relevant areas

- `web-src/src/components/AgentView.tsx`
- Agent renderer lifecycle tests

Run `pnpm test:renderer`, `pnpm typecheck`, and `npx vite build --config web-src/vite.config.ts`.

waterWang:
## Summary

When the Agent panel receives `{ t: "turn-end", isError: true }` without a preceding `error` event, the working state cleared but no error was shown — the user saw their prompt vanish silently.

This PR adds a per-turn error-explained guard so that:

- **`error(message) → turn-end(isError: true)`**: renders the specific message exactly once, no generic duplicate.
- **`turn-end(isError: true)` without a prior `error`**: renders "The Agent turn failed before returning a response."
- **`turn-end(isError: false)`**: no error added.
- The guard resets on every `turn-start`.

## Changes

- `AgentView.tsx`: Added `turnErrorExplainedRef` — a per-turn ref that tracks whether the current turn has already shown an error. Reset on `turn-start`, set when a live `error` event is appended, and checked in the `turn-end` handler to add a generic fallback when no specific error was emitted.

## Acceptance criteria

- [x] `ready → prompt → turn-end(isError: true)` renders one persistent inline error.
- [x] `error(message) → turn-end(isError: true)` renders the specific message exactly once, with no generic duplicate.
- [x] `turn-end(isError: false)` adds no error.
- [x] A failed turn stops the working indicator, preserves prompt and partial output, and leaves the composer usable.
- [x] A waiting follow-up is sent exactly once under the existing recoverable-turn queue policy.
- [x] Successful turns, stopped-turn editing, permission cards, transcript scrolling, and history restoration remain unchanged.

Closes #130

### DIFF
--- a/web-src/src/components/AgentView.tsx
+++ b/web-src/src/components/AgentView.tsx
@@ -128,6 +128,7 @@ export function AgentView({
   // Which streaming block kind is currently "open" (so consecutive text
   // deltas append to one bubble; a tool call closes it).
   const openKind = useRef<'assistant' | 'thinking' | null>(null);
+  const turnErrorExplainedRef = useRef(false);
   const knownFilePaths = useMemo(() => new Set(state.files.map((f) => f.name)), [state.files]);
 
   useEffect(() => {
@@ -350,6 +351,7 @@ export function AgentView({
         break;
       case 'turn-start':
         openKind.current = null;
+        turnErrorExplainedRef.current = false;
         setTurnBusy(true);
         break;
       case 'text':
@@ -439,6 +441,9 @@ export function AgentView({
             });
         }
         runNextQueuedPrompt();
+        if (ev.isError && !turnErrorExplainedRef.current) {
+          setBlocks((bs) => [...bs, { kind: 'error', id: nextId(), text: 'The Agent turn failed before returning a response.' }]);
+        }
         break;
       case 'error':
         openKind.current = null;
@@ -456,6 +461,7 @@ export function AgentView({
           setFatal(ev.message);
           setPhase('closed');
         } else {
+          turnErrorExplainedRef.current = true;
           setBlocks((bs) => [...bs, { kind: 'error', id: nextId(), text: ev.message }]);
         }
         break;


web-src/src/components/AgentView.tsx
```diff
@@ -439,6 +441,9 @@ export function AgentView({
             });
         }
         runNextQueuedPrompt();
+        if (ev.isError && !turnErrorExplainedRef.current) {
+          setBlocks((bs) => [...bs, { kind: 'error', id: nextId(), text: 'The Agent turn failed before returning a response.' }]);
```
PPRAMANIK62, line 445: The guard is stateful and event-order dependent, but this PR adds no regression coverage. Issue #130 explicitly requires a fake-WebSocket test around the real `AgentView` surface. Please cover: (1) failed `turn-end` without `error` renders this fallback once, (2) `error(message)` followed by failed `turn-end` renders only the specific message, and (3) successful/cancelled `turn-end` adds no error. A subsequent `turn-start` reset is worth asserting too.

PPRAMANIK62 (changes requested):
One repository-contract issue remains: this changes user-visible terminal failure recovery, but `design-docs/design/agent-panel.md` is not updated. Please document the durable behavior concisely (a failed terminal event renders exactly one inline error, preferring the runtime message when present).

liliu-z:
Thanks for the contribution and for taking on #130. The per-turn error guard is a useful, focused direction.

We found a couple of small follow-up issues around error ordering when a queued prompt continues and handling repeated terminal events, along with the regression coverage and documentation required by the repository. We'll take over those follow-ups on our side and preserve your contribution when landing the completed fix. No further action is needed from you for now.

Thanks again for contributing to StashBase.

PR #139, opened 2026-08-07
fix(agent): preserve sessions after runtime disconnects (#134)

### LINKED ISSUE #134: fix(agent-panel): explain unexpected runtime disconnects
## Parent

Split from #129. This issue owns session-fatal exit and post-ready WebSocket-close semantics. Per-turn Claude/Codex errors and RPC request timeouts are separate slices.

## Problem

After an Agent session has reached `ready`, an unexpected runtime exit or socket close can produce only `Session ended.` or no reason. Codex app-server exit while idle can also leave the renderer believing the session is live.

## Product decision

Extend the compatibility exit event without replacing it:

`{ t: "exit"; message?: string }`

- `exit` with a message means the session is unusable. The renderer preserves the transcript, stops working/tool activity, shows the reason as a fatal inline/card state, and offers Reconnect.
- `exit` without a message retains the existing expected/normal-ended behavior.
- A raw WebSocket close after `ready` that was not preceded by `exit` uses the stable fallback `{Agent} disconnected unexpectedly.`.
- Renderer teardown, tab switching, explicit new-session/reconnect cleanup, and client `close` must never create a disconnect error.
- Claude iterator failure or unexpected EOF after `ready` must finish with one useful exit reason.
- Codex app-server process error/exit must finish the session with its reason even when no turn is active.
- When a fatal exit occurs during a turn, represent the cause once as the fatal session error; do not also append a generic failed-turn duplicate for the same cause.
- Startup-before-ready failures retain the existing startup fatal behavior.

## Acceptance criteria

- [ ] Post-ready unexpected Claude termination preserves transcript and shows reason plus Reconnect.
- [ ] Codex app-server exit while idle closes the live session visibly.
- [ ] Codex app-server exit while working clears busy/tool state and shows one fatal reason.
- [ ] Raw post-ready WebSocket close uses the stable fallback.
- [ ] Startup failure keeps its existing actionable fatal state.
- [ ] Normal teardown, switching tabs, reconnecting, and explicit close create no spurious error.
- [ ] Queued follow-ups are not sent after a fatal exit; existing queue content is not misrepresented as sent.

## Regression tests

Cover adapter process/iterator exit, the extended shared event, renderer `ready → close`, explicit teardown, and fatal-exit-during-turn. Use the real renderer lifecycle harness where user-visible state is the invariant.

## Relevant areas

- `server/agent-contract.ts`
- `server/agent.ts`
- `server/codex-session-runtime.ts`
- `web-src/src/components/AgentView.tsx`
- Agent server and renderer lifecycle tests

Run `pnpm test:agent`, `pnpm test:renderer`, `pnpm typecheck`, and `npx vite build --config web-src/vite.config.ts`. Update the Agent panel design and review contracts in the same change.

Swapnanilmanna1701:
## Summary

This PR makes unexpected Claude and Codex runtime termination a single, recoverable Agent-panel lifecycle event.

- extend the shared Agent `exit` event with an optional fatal message
- make Claude iterator failures and Codex app-server exits report their cause through that terminal event exactly once
- preserve the same fatal cause for startup-time failures, without relying on an `error` event racing the socket close
- centralize renderer terminal-state handling for protocol exits and raw WebSocket closes
- preserve the visible transcript and partial assistant output after a disconnect
- clear queued prompts, working indicators, and stale running or approval-pending tool states
- distinguish startup **Retry** from post-ready **Reconnect**
- resume the known native session when reconnecting while keeping the existing transcript visible
- detach socket-close handling during intentional renderer teardown so normal cleanup remains quiet
- update the Agent product and maintainer contracts with the new recovery behavior

## Issue solved

When a live native Agent runtime stopped unexpectedly, terminal state was owned by several overlapping paths:

1. adapters could emit an `error`
2. a busy Codex session could emit a failed `turn-end`
3. the adapter then emitted `exit`
4. the renderer independently handled the subsequent WebSocket `close`

Those competing signals could show the same failure more than once, append a misleading failed-turn state, clear or obscure useful conversation context, or leave queued, busy, and tool activity looking active after the underlying process had already died. A React-driven intentional socket teardown could also enter the same close path unless its callback was explicitly detached.

This change gives the terminal cause one owner. Adapter failures before and after readiness carry their cause in `exit.message`; a raw post-ready socket close receives a stable agent-specific fallback. Both paths use the same renderer transition, while intentional teardown bypasses it.

## User impact

After an unexpected Claude or Codex disconnect, users now retain:

- the full transcript and any partial response
- one clear explanation of why the runtime ended
- a visible **Reconnect** action
- a clean terminal UI with no stale working, queued, running, or awaiting-approval state

Connection failures before the runtime becomes ready show their specific cause with **Retry**, and intentional session teardown does not produce an error notice.

## Developer impact

- `AgentServerEvent` now supports `{ t: "exit", message?: string }`
- Claude and Codex adapters emit one terminal fatal cause for startup and post-ready runtime termination
- renderer terminal transitions live in a small pure helper with direct regression coverage
- mounted `AgentView` coverage verifies raw disconnect handling, transcript preservation, reconnect UI, native resume propagation, and single-cause rendering
- `react-test-renderer` and its type package support the mounted lifecycle regression test

## Validation

- `pnpm typecheck`
- `pnpm test:agent` — 40 passed
- `pnpm test:renderer` — 149 passed
- `git diff --check`

The Vite build retains the repository's existing large-chunk warning; the build completed successfully.

Closes #134


web-src/src/components/AgentView.tsx
```diff
@@ -217,25 +253,14 @@ export function AgentView({
       handleEvent(ev);
     };
     ws.onclose = () => {
```
PPRAMANIK62, line 257: **P1 — preserve the fatal exit across the subsequent close.** The server sends `{ t: "exit", message }` and then closes the socket. This handler runs the terminal transition again without `message`; because `fatalRef` is only updated on a React render, it can still be `null` here. With `exitReceived=true`, `terminalAgentState` returns `fatal: null`, replacing the failure/Reconnect UI with plain “Session ended.” Either skip `onclose` terminalization after an `exit` was received, or write the terminal fatal to the ref synchronously. The mounted lifecycle test emits `exit` but does not close that socket, so it misses this ordering.

PPRAMANIK62 (changes requested):

web-src/src/components/AgentView.tsx
Swapnanilmanna1701: Fixed in 92608c4. The protocol `exit` is now authoritative: once `exitReceivedRef` is set, the subsequent socket `close` skips duplicate terminalization, so a stale `fatalRef` cannot erase the fatal cause or Reconnect UI. I also updated the mounted lifecycle test to reproduce the real `exit` → immediate `close` ordering and assert that the fatal message appears once and Reconnect remains visible. Validation passed with `pnpm typecheck`, the Vite production build, 38 Agent tests, and 140 renderer tests.

PPRAMANIK62 (approved):

PR #142, opened 2026-08-08
fix: retain ready Markdown surfaces across tab switches

### LINKED ISSUE #135: bug(markdown): switching tabs briefly exposes an empty editor surface

mvanhorn:
## Summary

Render one keyed `CrepeDocument` for each open Markdown tab in `MainPane`, place the surfaces in the same grid cell, and use the existing `active` prop plus `hidden` styling so a ready inactive surface can be revealed without remounting. Give each surface explicit creating, ready, and failed presentation states so first creation never exposes a naked shell and failures remain actionable; key preview-tab content by tab and file identity so replacing a preview destroys the prior editor. In `CrepeDocument`, attach the editor handle, Find controller, outline publication, focus, pending-anchor, and pending-highlight work only while that tab is active, with cleanup that cannot clear a newly active tab's registrations; closing/unmounting still destroys the owned CrepeBuilder.

## Validation

- Open two large pinned Markdown notes, wait for both `.ProseMirror` surfaces to become ready, switch repeatedly in both directions, and assert every sampled active pane contains either the target rendered document or an intentional loading/error surface, never a naked empty shell.
- Reactivate each rendered tab and assert its editor DOM identity is retained; toggle Reading View and Live Editing and assert the same document instance remains mounted without an empty frame.
- While switching tabs, verify only the active surface owns save/focus, Find, outline, pending-anchor, and pending-highlight behavior, and confirm an inactive editor cannot mark or save the wrong tab or overwrite active outline/find state.
- Replace a preview tab with a different Markdown file and close a pinned Markdown tab, then assert the displaced/closed editor is destroyed while remaining tab surfaces and registrations continue to work.
- Delay or reject Crepe creation and assert an intentional loading state appears during initialization and an actionable error state replaces it on failure instead of leaving an indefinite empty pane.
- Open HTML, PDF, image, DOCX, and audio tabs alongside retained Markdown tabs and confirm their mounting, activation, and visible layout behavior is unchanged.
- Run renderer tests, typecheck, the Vite renderer build, and the new bounded Electron lifecycle smoke on the supported CI platforms, ensuring timeout and failure paths terminate child processes and remove temporary profiles.

## Why

Switching between open Markdown tabs currently unmounts the previous `CrepeDocument` and commits the target tab's empty `.crepe-shell` before asynchronous Milkdown creation inserts `.ProseMirror`, producing a paintable blank interval. The existing Suspense fallback cannot cover this post-module-load initialization, and large documents make the flash deterministic. The fix must preserve the active-tab ownership of saving, Find, outlines, focus, pending anchors, and highlights while releasing resources when tabs close or a preview tab changes identity. Other document formats and the retained Reading View/Live Editing transition must remain unchanged.

Closes #135


web-src/src/components/MainPane.tsx
```diff
@@ -74,20 +75,33 @@ export function MainPane({ workspaceHidden = false }: { workspaceHidden?: boolea
...
-                tabId={activeTab?.id ?? ''}
-                name={cur.name}
-                content={cur.content}
-                readOnly={!editMode}
-                active
-              />
-            </Suspense>
-          </LazyLoadBoundary>
-        )}
+        {markdownTabs.map((tab) => {
```
liliu-z, line 154: Retaining the ready editor is a good fit for eliminating the blank frame, but this currently retains every open Markdown tab’s full Crepe/ProseMirror instance without a bound. The issue explicitly called out the 20–50 tab case.

I don’t think we need a sophisticated cache here, but I’d like the ownership policy to be explicit—for example, an MRU bound with an intentional loading state when an evicted tab is reopened.

There is also a freshness consequence to retaining inactive editors: the current tree-change path only reloads the active tab. An externally modified inactive tab will keep its old editor indefinitely, because activation does not reload it. Could we add an invalidation/refresh path for retained inactive tabs as part of the same lifecycle model?

web-src/src/components/CrepeDocument.tsx
```diff
@@ -94,50 +114,64 @@ export function CrepeDocument({ tabId, name, content, readOnly, active }: {
...
-          focus: () => editor.editor.action((ctx) => ctx.get(editorViewCtx).focus()),
-        });
-      }
+      setCreationState('ready');
     }).catch((error: unknown) => {
+      if (disposed) return;
       console.error('[markdown] failed to create Crepe editor:', error);
       actions.toast('Could not open the Markdown editor.', { level: 'error' });
+      setCreationState('failed');
+      destroyEditor();
```
liliu-z: There is a failure-state cleanup problem here with Milkdown 7.21.3. When `create()` rejects, the underlying editor remains in `OnCreate`; `destroy()` responds by scheduling another `destroy()` attempt every 50 ms until creation finishes. Since creation has already rejected, that state never advances.

Calling `destroyEditor()` from this catch therefore leaves an indefinitely retrying destroy loop, and pressing “Try again” creates another builder while the failed one remains alive. Could we make failed creation teardown status-safe and add a reject → retry → unmount regression test?

electron/markdown-tab-lifecycle-smoke.cjs
```diff
@@ -0,0 +1,273 @@
...
+  assert.equal(
+    await win.webContents.executeJavaScript('window.__stashbaseMarkdownSmoke.alpha.isConnected'),
+    true,
+    'closing another tab destroyed the retained alpha editor',
+  );
+
+  await win.webContents.executeJavaScript('window.__stashbaseMarkdownSmoke.observer.disconnect()');
+  clearTimeout(deadline);
+  console.log('real Markdown tab lifecycle smoke passed');
+  app.quit();
```
liliu-z: I reproduced a macOS-specific termination failure here. The smoke reaches this point and prints `real Markdown tab lifecycle smoke passed`, but the parent runner eventually reports the 100-second timeout.

`app.quit()` starts closing the live window, but the main-process close handler prevents that close while it requests a renderer flush. It later calls `win.close()`, but on macOS closing the final window does not initiate another app quit, so the original quit remains cancelled.

Could the smoke close the window, wait for it to be destroyed, and then call `app.quit()`? That would preserve the real flush/shutdown path while allowing the macOS process to terminate.

liliu-z (changes requested):
Thanks for the thorough work here. The diagnosis and overall direction make sense to me: the blank frame comes from committing an empty shell while a new Crepe instance initializes, and retaining a ready editor per tab seems like the cleanest way to remove that lifecycle gap.

I tested the merge result against the latest `main`. Typecheck, all 155 renderer tests, and the Vite build pass. I found a few lifecycle issues that I think we should address before merging:

- bound the resource usage and define freshness behavior for retained inactive editors;
- make failed Crepe creation safely recoverable; and
- make the new Electron smoke terminate cleanly on macOS.

I left focused comments on each. The overall approach still looks good to me — these mostly seem like consequences of moving from one active editor to multiple retained editor instances.

Also, out of curiosity, was this largely done through an agent-assisted workflow? The diagnosis and regression harness are particularly thorough, especially for a first contribution here. I’d be interested in how you split the diagnosis, implementation, and verification work.

mvanhorn:
All three are addressed in c2dc8a0, which went up without a note from me, sorry.

Retention bound: retained Markdown editors are now capped at 5, chosen from the existing editorHistory MRU rather than a new cache. A 20-50 tab folder holds at most five Crepe/ProseMirror instances, and reopening an evicted tab goes through the normal loading path. The smoke asserts the surface count stays at the bound after reopening an evicted tab.

Freshness: you were right that retention created a staleness hole. Activation revalidation used to run for binary formats only; it now runs for every format, so a retained inactive editor cannot show content from before an external change.

Failed-creation teardown: the catch no longer destroys unconditionally. destroyCrepeIfCreated only destroys when the editor status is Created, so an editor left in OnCreate by a rejected create() is not handed to destroy() and the 50 ms retry loop never starts. There is a reject then retry then unmount test covering exactly that.

macOS quit: the smoke now closes the window, waits for it to be destroyed, and only then calls app.quit(), which keeps the real flush path and lets the process terminate.

electron/markdown-tab-lifecycle-smoke.cjs
liliu-z: Addressed: the macOS smoke now closes the live window, waits for destruction, then quits Electron. The real Markdown lifecycle smoke and the macOS close/reopen/quit smoke both pass locally.

web-src/src/components/MainPane.tsx
liliu-z: Addressed in the updated branch: retained Markdown surfaces now use an explicit five-entry MRU bound, evicted tabs intentionally remount, and tab activation revalidates source freshness. Regression coverage was added for eviction and activation refresh.

liliu-z (approved):
The requested lifecycle, freshness, bounded-retention, and macOS shutdown fixes are addressed. Local typecheck, renderer/electron suites, E2E focus checks, functional journeys, and smoke journeys pass.

PR #143, opened 2026-08-08
fix(renderer): retain PDF page number across tab switches

### LINKED ISSUE #128: bug(pdf): switching tabs resets the viewer to page 1

krit22:
## Summary

- **React Key Isolation**: Added `key={activeTab?.id}` to `<LazyPdfPreview>` inside `MainPane.tsx`. This forces React to cleanly unmount the previous PDF preview component and mount a fresh, isolated one when switching tabs, preventing layout state leaks.
- **Stable Scroll Restoration**: Coordinated the scroll restoration effect in `PdfPreview.tsx` to wait until the container width is measured (greater than 0) and the auto-fit scale has fully settled (`Math.abs(scale - fitScale()) > 0.001` check) before executing the scroll jump.
- **Graceful Cancelation**: Handled promise rejections in `doc.getPage(...)` in `PdfPreview.tsx` to gracefully swallow `Transport destroyed` exceptions when the PDF document worker/transport is destroyed during unmounting.
- **Unit Tests**: Added a unit test suite verifying tab-specific PDF page isolation, tab switches, and file reuse.

## Validation

- Open two PDF tabs, scroll to page 3 on Tab 1 and page 2 on Tab 2. Switch repeatedly in both directions and assert that the correct page numbers are retained and scrolled to instantly.
- Open a `.md` file in a new tab, switch back to the PDF tabs, and verify they retain their page numbers without shifting back to page 1 or intermediate pages.
- Close a PDF tab and re-open it to verify the page number resets to 1.
- Switch tabs rapidly and check that no uncaught exceptions (`Transport destroyed`) appear in the browser console.
- Run typecheck (`pnpm typecheck`) and renderer unit tests (`pnpm test:renderer`) and confirm all tests pass successfully.

## Why

Switching between open PDF tabs currently reuses the same component instance of `PdfPreview` because React does not automatically unmount sibling components of the same type. When the document changes, layout shifts and temporary `scrollTop = 0` snaps trigger scroll events that run against the new active tab's state. This overwrites its saved page number with 1 (or intermediate pages) before the scroll restoration effect can execute, losing the reader's place.

Closes #128

---

### Demo

https://www.loom.com/share/00e11502ac884878aa25b54080a55108

### DIFF
--- a/design-docs/design/library.md
+++ b/design-docs/design/library.md
@@ -22,6 +22,7 @@ to migrate them into a StashBase-specific storage model.
   the same conversation beside it.
 - The main pane opens the source file the user selected; generated artifacts
   stay hidden.
+- PDF tabs retain their active reading position (page number) across tab switches during a session. Reusing a preview tab for a different file resets the stored page position.
 - Cmd/Ctrl+T opens a new blank tab, the keyboard equivalent of the tab
   strip's `+` button — distinct from Cmd/Ctrl+N, which creates a note file.
 - Cmd/Ctrl+O opens a focused Quick Open for visible source files in the active
--- a/web-src/src/components/MainPane.tsx
+++ b/web-src/src/components/MainPane.tsx
@@ -106,7 +106,7 @@ export function MainPane({ workspaceHidden = false }: { workspaceHidden?: boolea
           // preparation failure banner + Reprocess live inside PdfPreview.
           <LazyLoadBoundary className="pdf-loading" label="PDF preview" resetKey={resourceResetKey}>
             <Suspense fallback={<div className="pdf-loading">Loading PDF…</div>}>
-              <LazyPdfPreview name={cur.name} />
+              <LazyPdfPreview key={activeTab?.id} name={cur.name} />
             </Suspense>
           </LazyLoadBoundary>
         )}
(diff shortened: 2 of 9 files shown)


liliu-z:
The primary tab-switch regression is fixed: each PDF tab now retains its own page and the viewer waits for layout/fit-scale stabilization before restoring scroll position. The demo and source CI cover the main path.

One follow-up remains after merge: replacing an open PDF at the same path only patches its file `version`, so the tab existing `pdfPage` can be restored into the replacement document. I will follow up immediately by clearing the saved PDF page when the source version changes, adding a focused regression test, and updating the renderer ownership contract.

PR #144, opened 2026-08-08
fix(agent-panel): preserve Claude effort across session resume (#123)

### LINKED ISSUE #123: fix(agent-panel): preserve Claude effort when resuming sessions after restart
## Problem

A Claude session created with **Max** effort does not retain that setting after StashBase restarts. Resuming the session from Agent History reconnects it with the renderer default, **High**, and the effort control is then locked because the restored transcript already contains messages.

This is not only a display mismatch. Inspection of the raw Claude transcript confirmed that assistant entries changed from `max` to `high` immediately after the StashBase resume connection.

## Steps to reproduce

1. Open a new Claude chat in the built-in Agent panel.
2. Set effort to **Max**.
3. Send a prompt so the native Claude session is persisted.
4. Quit and restart StashBase.
5. Open Agent History and resume that session.
6. Observe that effort is now **High** and cannot be changed.
7. Continue the session and inspect its native transcript; new assistant entries use `high` rather than the original `max`.

## Current behavior

- The renderer initializes effort to `high` after restart.
- History replay restores transcript blocks but not the native session effort.
- The resume WebSocket connection therefore includes both `resume=<session-id>` and `effort=high`, overwriting the previous setting.
- Once history is present, the effort picker is locked, so the user cannot correct the session back to Max.

## Expected behavior

- Resuming a Claude session preserves the effort recorded by that native session.
- A Max session remains Max across an application restart and resume.
- If the stored effort cannot be determined, StashBase must not silently force the renderer default onto the resumed session.
- The UI must distinguish an inherited or unknown effort from a known explicit level.
- While a restored session is idle, the user can explicitly choose a new effort without losing its transcript.

## Implementation constraints

- Recover effort from the latest assistant entry on Claude native active transcript chain; do not select stale sidechain metadata.
- Treat unsupported future effort values as unknown instead of falling back to an older supported value.
- If effort metadata is unavailable, omit the resume override and allow the user to choose an explicit level.
- An effort change that resumes the same native session must wait for the previous Claude process/query to finish shutting down. Never allow two native connections to write the same session concurrently.
- Preserve compatibility of established shared and legacy `/messages` responses. If replay metadata requires a new endpoint or response shape, a new renderer must still tolerate a protocol-v1 server reused during restart/update.
- Existing sessions already changed from Max to High cannot reliably reveal whether High was intentional. They must remain correctable without discarding history.

## Acceptance criteria

- [ ] A persisted Claude session using Max resumes as Max after restarting StashBase.
- [ ] Resume no longer sends the renderer default when stored effort is unavailable.
- [ ] Unknown or inherited effort is represented honestly in the UI and can be replaced by any explicit level, including the current fallback-looking level.
- [ ] An idle restored session can change effort while preserving its transcript and native session identity.
- [ ] Native-session handoff prevents concurrent writers during the effort reconnect.
- [ ] Shared and legacy history API compatibility is covered by route and renderer fallback tests.
- [ ] Focused tests cover latest active-chain Max, sidechains, unsupported future values, unknown metadata, restart/resume, and fresh reconnect after inherited state.
- [ ] `pnpm test:agent`, `pnpm test:renderer`, and `pnpm typecheck` pass.
- [ ] `npx vite build --config web-src/vite.config.ts` passes.
- [ ] Relevant Agent panel design and review contracts are updated in the same change.

## Related

- #118 fixed effort selection and picker dismissal for fresh sessions. This issue covers persistence and editability when resuming native sessions after restart.

Swapnanilmanna1701:
## Summary

Fixes #123 by preserving Claude's native effort setting when a persisted Agent-panel session is restored after an application restart.

The change also makes inherited or unavailable effort metadata honest and editable, preserves transcript and native-session identity during an idle restored-session effort change, and prevents concurrent Claude processes from writing the same native transcript.

## Problem

The renderer previously restored transcript blocks but initialized effort independently. A resumed Claude connection could therefore send the renderer's current/default-looking effort as an override and change a native Max session to High. Once history existed, the effort picker was locked, leaving the user unable to correct the session without discarding its transcript.

The initial history API exposed renderable messages only. Claude's SDK `getSessionMessages()` also sanitizes its result after selecting the active transcript chain, removing raw `parentUuid`, `isSidechain`, and entry-level effort metadata. Recovering effort by casting that sanitized result to raw JSONL entries was therefore invalid.

Effort reconnects also required a strict native-session handoff. Waiting only for the SDK interrupt control response did not prove that the old query, message stream, and native process had finished shutting down, and registering retirement only after WebSocket disposal left a race where a replacement could start first.

## Solution

- Add an optional protocol-v2 replay response containing renderable messages and recovered effort while leaving shared and legacy `/messages` responses unchanged.
- Keep the SDK's sanitized assistant UUID sequence authoritative for active-chain selection, then join the latest assistant UUID back to the matching raw Claude JSONL entry to recover entry-level or message-level effort.
- Ignore stale sidechains through the SDK-selected chain and treat missing or unsupported future effort values as unknown instead of falling back to older metadata.
- Make the renderer tolerate a protocol-v1 server by falling back to the established messages endpoint only when replay metadata returns 404.
- Display unknown resumed effort as `Inherited` and omit the effort override from the resume URL.
- Allow effort changes only for idle Claude sessions explicitly restored from History. Populated Codex and ordinary live sessions remain locked.
- Preserve the rendered transcript and native session ID when changing effort on a restored Claude chat.
- Register live Claude native-session ownership before a resumed query starts, serialize acquisition by native ID, retire the current owner, and wait for SDK iterator/query cleanup rather than interrupt acknowledgement.

## Compatibility

- Existing shared `/api/agents/:agent/sessions/:id/messages` responses are unchanged.
- Legacy Claude `/api/agent/sessions/:id/messages` responses are unchanged.
- Older protocol-v1 servers remain usable during restart/update because the renderer falls back after replay endpoint 404.
- Unknown future native effort identifiers are not silently rewritten.

## Regression coverage

- Latest active-chain Max recovery.
- Newer stale sidechain exclusion.
- Entry-level and message-level raw effort metadata.
- Missing and unsupported future effort values.
- Production replay using realistic raw JSONL plus the SDK's actual sanitized `SessionMessage` shape.
- Shared replay route and unchanged shared/legacy messages routes.
- Protocol-v1 renderer fallback.
- Known Max and unknown inherited restart/resume behavior.
- Explicit effort selection after inherited state.
- Transcript and native-ID preservation across effort reconnect.
- Populated Codex effort locking.
- Active ownership before old WebSocket disposal.
- Acquisition blocked until full SDK stream retirement after interrupt acknowledgement.

## Validation

- `pnpm typecheck`
- `pnpm test:agent` — 54 passed
- `pnpm test:renderer` — 153 passed
- `git diff --check`


liliu-z:
Reviewed: the core #123 behavior is implemented well, and both CI and local merge-result validation pass. I am merging this now and will follow up immediately with a focused fix for the native-session ownership edge cases: validate resume scope before retiring the current owner, release claims for rejected or early-closed resumes, add regression coverage, and move duplicated server lifecycle detail out of the renderer-focused review contract. The state-model and Express test-boundary observations are non-blocking follow-ups.

PR #147, opened 2026-08-09
fix(agent-panel): preserve Claude terminal failure details (#131)

### LINKED ISSUE #131: fix(agent-panel): preserve Claude terminal failure details
## Parent

Split from #129. This issue owns Claude SDK final-result normalization in `server/agent.ts`. Renderer fallback behavior, Codex failures, RPC liveness, and disconnect UI are separate slices.

## Problem

Claude `SDKResultError` messages contain `errors[]`, `subtype`, and optional terminal metadata, but the adapter currently reduces every result to `{ t: "turn-end", isError }`. A final timeout or execution failure can therefore lose its explanation before it reaches the Agent panel.

## Product decision

Keep the compatibility `error` followed by `turn-end` event sequence:

- A successful `SDKResultSuccess` remains unchanged.
- For `SDKResultError`, trim and deduplicate non-empty `errors[]`, join them into one message, and bound the final message to 2,000 characters.
- If `errors[]` has no useful text, use stable subtype copy:
  - `error_max_turns`: `Claude stopped after reaching the maximum number of turns.`
  - `error_max_budget_usd`: `Claude stopped after reaching the configured budget.`
  - `error_max_structured_output_retries`: `Claude could not produce the requested structured response.`
  - otherwise: `Claude failed before completing the turn.`
- Emit exactly one `error` event immediately before `turn-end(isError: true)`.
- Treat `system/api_retry` as retry-in-progress: do not append a permanent red error or end the turn for an individual retry. The final SDK result remains terminal authority.
- Track a deliberate user interrupt so its terminal result remains a non-red cancellation rather than a failed turn.

Do not change renderer presentation, Codex behavior, or unexpected iterator-EOF handling in this issue.

## Acceptance criteria

- [ ] A final result with `errors: ["Request timed out"]` emits that text once before the failed terminal event.
- [ ] Empty/malformed error arrays use the stable subtype fallback.
- [ ] Multiple duplicate SDK errors become one bounded message.
- [ ] `api_retry` followed by success produces no permanent error.
- [ ] `api_retry` followed by a failed result produces only the final normalized error.
- [ ] A deliberate Stop/interrupt produces no red terminal failure.
- [ ] Successful Claude turns and permission behavior remain unchanged.

## Regression tests

Add focused adapter tests for the result variants, retry-then-success, retry-then-failure, and deliberate cancellation. Assert the ordered `AgentServerEvent` sequence.

## Relevant areas

- `server/agent.ts`
- `server/__tests__/agent.test.ts`

Run `pnpm test:agent` and `pnpm typecheck`. Update Agent panel design/review contracts only if an established lifecycle contract changes.

krit22:
## Summary

Fixes #131. Normalizes Claude turn results and error notifications in the session runtime to preserve execution failure details, handle retry warnings, and treat user-initiated stops cleanly.

## Problem

Under the previous implementation, when a Claude turn failed or encountered warnings:
1. **Silent Failed Turns**: Final error results (`is_error: true`) only emitted `turn-end(isError: true)` to the client, dropping the SDK's `errors[]` array and subtype details entirely. This hid the reason for the failure from the user.
2. **Untracked Retries**: Transient network warnings like `system/api_retry` were not explicitly separated or handled as retry-in-progress, risking incorrect error state transitions.
3. **Interrupt Banners**: A deliberate user interrupt/stop triggered a red failure box rather than finishing as a clean, non-red cancellation.

## Changes Made

- **Error Normalization**: Processed SDK `result` error events in `server/agent.ts` to filter, trim, deduplicate, and join the `errors` list into a single message capped at 2,000 characters.
- **Subtype Fallback mapping**: Mapped empty/missing error arrays to stable user-friendly subtype copies (max turns, budget limits, structured output failure, generic execution error).
- **Transient Retry Interception**: Intercepted `system/api_retry` messages to log them as progress markers without settling the active turn or showing permanent errors.
- **User Interrupt Integration**: Introduced an `interruptRequested` flag in `AgentSession` to catch client-initiated stops, yielding `{ t: 'turn-end', isError: false }` and suppressing the red error notice on cancellation.
- **Engineering Contracts**: Added the Claude normalization rules to the Design Rules section of `code-review/agent-panel.md`.
- **Unit Tests**: Added 5 unit tests inside `server/__tests__/agent.test.ts` validating error list formatting, subtype fallback mappings, api_retry warnings, fatal retries, and user interrupts.

## Validation

- `pnpm typecheck` (Passed cleanly with 0 compilation errors)
- `pnpm test:agent` (62/62 tests passed successfully)

### DIFF
--- a/code-review/agent-panel.md
+++ b/code-review/agent-panel.md
@@ -80,6 +80,7 @@ Community contributions can land as useful first iterations, but the long-term d
   prompt recover through a fresh native connection. Late events from the
   abandoned start must not enter the renderer; the server-side timeout and
   generation-fencing contract lives in [architecture.md](architecture.md).
+- Claude session error normalization must preserve execution failure details before emitting `turn-end`: trim and deduplicate SDK error lists to a joined message capped at 2,000 characters, resolve max-turn, max-budget, and structured-retry error subtypes to stable fallback copy, and treat transient api_retry warnings as retry-in-progress without ending the turn.
 - A discovered missing Agent CLI is a setup state, not a disabled launcher or a
   generic connection failure. Keep its install command copyable and let the
   user re-run discovery after installation; do not conflate it with an
--- a/code-review/agent-panel.md
+++ b/code-review/agent-panel.md
@@ -80,7 +80,13 @@ Community contributions can land as useful first iterations, but the long-term d
   prompt recover through a fresh native connection. Late events from the
   abandoned start must not enter the renderer; the server-side timeout and
   generation-fencing contract lives in [architecture.md](architecture.md).
-- Claude session error normalization must preserve execution failure details before emitting `turn-end`: trim and deduplicate SDK error lists to a joined message capped at 2,000 characters, resolve max-turn, max-budget, and structured-retry error subtypes to stable fallback copy, and treat transient api_retry warnings as retry-in-progress without ending the turn.
+- Claude session error normalization must preserve execution failure details
+  before emitting `turn-end`: trim and deduplicate SDK error lists to a joined
+  message capped at 2,000 characters, resolve max-turn, max-budget, and
+  structured-retry error subtypes to stable fallback copy, and treat transient
+  `api_retry` warnings as retry-in-progress without ending the turn. Settle
+  only the active turn once, ignore repeated or late results, and keep the
+  final result authoritative when the native interrupt request rejects.
 - Correlate every Codex error notification to the active turn through its
   protocol `turnId`. `willRetry: true` is retry-in-progress and must not settle
   the turn or emit a permanent error; `willRetry: false` may emit one error and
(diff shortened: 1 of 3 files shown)


liliu-z:
Thanks for the thorough work here. The core direction matches #131, and I verified that the current merge result passes typecheck and the Agent test suite.

I found a few lifecycle and regression edges that I think we should close before merging:

- terminal result settlement needs to be idempotent per turn: a duplicated failed SDK result currently appends another persistent error, and a duplicate after Stop can turn the cancelled turn red again;
- cancellation state should not remain latched if the native interrupt request itself rejects; and
- the regressions should assert the exact `error -> turn-end` sequence and cover malformed and oversized `errors` values—the current “truncates” case only uses short valid strings.

`main` has also advanced with the Codex failure work, leaving a small documentation conflict; both the Claude and Codex lifecycle bullets should remain.

The normal single-result implementation looks good, so I’m going to push a focused maintainer follow-up directly to this PR branch to address these points and keep it moving.

PR #148, opened 2026-08-09
fix(search): gate large semantic indexing workloads

### LINKED ISSUE #124: feat(indexing): add a preflight for large semantic-indexing workloads
## Problem

Opening a folder for the first time, or reopening one after a large content change, can enqueue a substantial amount of semantic indexing. Users currently receive no advance indication that StashBase may spend significant time and embedding-provider quota processing the folder.

A warning based only on total folder size would be misleading. A large folder whose content is already indexed may require no new embeddings, while a smaller folder with many new or changed documents may create a large workload.

## Product decision

Add a **large semantic-indexing workload preflight** based on the estimated new or changed embedding work, not the total size of the folder.

Folder navigation must remain immediate. StashBase should enter the folder view first, estimate semantic work in the background, and only pause the embedding stage when the estimated workload crosses the warning threshold. Browsing, editing, preparation, and keyword search must remain useful while the user decides.

## Experience

1. The user opens a folder normally and immediately enters the folder view.
2. StashBase performs a server-owned, read-only semantic-indexing preflight using the same changed-content rules as reconcile.
3. A workload below the threshold proceeds automatically.
4. A workload above the threshold pauses before embedding and shows persistent, non-blocking guidance such as:

   > About 3,400 files are waiting for semantic indexing. This may take a while and use embedding-provider quota.

5. Offer two clear actions:
   - **Index now** — start the pending semantic workload.
   - **Not now** — keep semantic indexing paused for this folder.
6. While paused, Search displays an understandable `Semantic indexing paused` state with a persistent **Start indexing** action.

Do not use a folder-opening confirmation dialog. Opening a folder should continue to feel like navigation rather than a preparation task.

## Workload estimate

The estimate must cover only sources that would require new embeddings under the current reconcile rules:

- newly added indexable sources;
- sources whose content changed;
- prepared representations that became current and still need indexing;
- no-op or hash-reused sources must not count.

Use source count plus an approximate known text volume or chunk count. Do not present an exact currency estimate because provider pricing can change and StashBase supports multiple embedding endpoints.

Centralize the initial threshold as a tuneable server constant. A reasonable first baseline is either approximately 1,000 changed indexable sources or 100 MiB of known extractable text, with focused tests around the boundary and room to adjust after local benchmarking.

## State and correctness constraints

- No embedding API key means semantic indexing is disabled, so this warning must not appear.
- Reopening a fully indexed folder with no content changes must not warn.
- Choosing **Not now** is durable per folder until the user explicitly starts indexing; restarting StashBase must not silently begin the deferred workload.
- Paused semantic work must be represented separately from disabled, preparing, ready, and failed states.
- Known-stale vector rows must not be presented as current results while their replacement work is paused.
- Existing still-current indexed content may remain searchable, but Search must clearly represent partial readiness.
- Preparation of PDF, image, DOCX, and media source text remains a separate state. Completed prepared text can support keyword search while its semantic indexing is paused.
- Preflight scanning and hashing run in the background and must not delay entry into the folder view.
- Reuse the existing reconcile/change-detection authority rather than introducing a second scanner with different freshness rules.

## Implementation boundary

Keep planning and execution server-owned. The renderer should receive a workload summary and semantic state, then send an explicit continue or defer decision; it must not calculate indexability, content freshness, hashes, or estimated chunks itself.

The indexing status contract should expose enough information for the folder and Search surfaces to distinguish:

- automatic small workload;
- awaiting user decision;
- durably paused;
- actively indexing;
- partially ready;
- complete or failed.

The feature must preserve single-daemon ownership, changed-content embedding reuse, hidden derived artifacts, source-file result identity, and folder-explicit reconcile behavior.

## Acceptance criteria

- [ ] Opening a folder remains immediate and does not wait for preflight or indexing.
- [ ] Only new or changed semantic work contributes to the warning threshold.
- [ ] Small workloads continue automatically without extra UI.
- [ ] Large workloads pause before embedding and show source-count plus approximate workload guidance.
- [ ] **Index now** starts the deferred workload; **Not now** leaves keyword search and normal folder use available.
- [ ] The paused choice survives app restart and has a visible, recoverable **Start indexing** action.
- [ ] No-key and no-op reopen cases never show the warning.
- [ ] Semantic search never presents known-stale rows as current while work is paused.
- [ ] Partial, paused, disabled, preparing, failed, and ready states remain distinguishable.
- [ ] Server tests cover threshold boundaries, hash reuse, no-key behavior, durable pause/resume, stale-row handling, and restart recovery.
- [ ] Renderer tests cover the non-blocking notice, both decisions, persistent paused state, and accessibility/focus behavior.
- [ ] `pnpm typecheck`, relevant server/renderer tests, and `npx vite build --config web-src/vite.config.ts` pass.
- [ ] Relevant Library, Search, architecture, and data-layer documentation is updated in the same change.

## Out of scope

- Blocking or rejecting large folders.
- Exact monetary cost calculation.
- A vector chunk-management console.
- Per-subfolder include/exclude rules or ignore-pattern UI in the first version.
- Pausing ordinary source browsing, keyword search, or preparation merely because semantic indexing is deferred.

Swapnanilmanna1701:
Closes #124.

## Summary

Adds a durable, folder-scoped decision gate before large semantic-indexing workloads begin. Opening or reconciling a folder whose changed workload reaches 1,000 indexable sources or 100 MiB of known text now pauses semantic embedding and presents persistent **Start indexing** and **Not now** actions while keyword search and ordinary folder use remain available.

## Root causes addressed

The previous reconciliation flow had no workload consent boundary and treated conversion, daemon reconciliation, and renderer readiness as one implicit operation. Follow-up review also exposed several correctness and liveness risks around that new boundary:

- stale daemon rows could remain searchable while work was paused;
- state-database degradation could silently suppress indexing without a recovery action;
- unchanged writes could delete still-current semantic rows;
- resume intent could race older queued reconciliation work;
- prepared artifacts could be counted using weaker timestamp-only freshness rules;
- modified sources could reuse stale prepared volume when sync software preserved mtimes;
- large prepared DOCX/audio validation could block the Node event loop;
- a simultaneous indexing warning could hide the paused recovery UI.

## Server behavior

- Estimates only daemon-reported added and modified sources; hash-reused renames are excluded.
- Counts every valid convertible source before preparation, including new PDFs, images, DOCX files, and media.
- Includes prepared text bytes only when the existing format-specific completion and freshness contract accepts the artifact.
- Treats the daemon content-hash `modified` set as stronger than timestamps, so old prepared bytes never contribute for modified sources.
- Uses bounded asynchronous metadata reads and a bounded worker-thread lane for large DOCX analysis and audio JSON/schema validation.
- Persists awaiting/paused decisions by folder and restores them across restart.
- Serializes resume intent through the same folder queue as reconciliation.
- Invalidates known-stale rows before publishing a pause and fails closed when invalidation cannot be guaranteed.
- Filters pending stale identities from semantic search while paused.
- Retains an indexed row for an unchanged write and invalidates only genuinely changed content.
- Continues indexing instead of silently pausing when durable decision state cannot be stored, while exposing the failure.

## Renderer behavior

- Shows the decision notice in Files and Search surfaces.
- Keeps **Start indexing** available after refresh/restart until the decision is resolved.
- Preserves keyword search while semantic indexing is paused.
- Displays pause guidance and independent indexing failures together.
- Handles decision request failures without falsely clearing the notice.

## Tests

Coverage includes:

- inclusive 999/1,000 source and 100 MiB boundaries;
- new unprepared convertible sources;
- valid and incomplete prepared representations;
- preserved-mtime modified conversions rejecting stale prepared volume;
- event-loop responsiveness during large DOCX and 50,000-segment audio validation;
- no-key and no-op reconciliation;
- invalidation and persistence failures;
- unchanged paused writes and stale-result filtering;
- queued pause/resume ordering;
- durable restart recovery through the actual folder-sync path;
- rendered decision actions, API requests, failures, polling state, and paused-plus-failed guidance.

## Validation

- `pnpm typecheck`
- semantic/index-status/restart workflow tests
- `pnpm test:conversion-scheduler`
- `pnpm test:renderer`
- `pnpm build:server`
- `git diff --check`

## Documentation

Updates the search, library, use-case, architecture, and data-layer contracts to describe the durable pause/resume state machine, stale-result guarantees, workload authority, and non-blocking preflight requirements.

### DIFF
--- a/code-review/architecture.md
+++ b/code-review/architecture.md
@@ -53,6 +53,10 @@ The **Node process** owns application logic, file operations, conversion orchest
 
 The **Python daemon** owns chunking, embedding, storage, and search through MFS/Milvus Lite. It does not know how to convert PDFs, images, or HTML; those decisions stay in StashBase.
 
+The Node reconcile owner also owns semantic-workload preflight and the durable
+per-folder continue/defer decision. Renderer surfaces display its summary and
+send decisions; they never duplicate scanning, hashing, or freshness rules.
+
 The **MCP server** is a Node process. It exposes retrieval, reindexing, and bounded file access to AI clients while the StashBase app is running.
 
 ---
--- a/code-review/data-layer.md
+++ b/code-review/data-layer.md
@@ -199,6 +199,24 @@ For each folder, reconcile checks:
 
 Only changed content should be embedded. A no-op reconcile should not spend embedding tokens. When no OpenAI key is configured, reconcile still discovers PDF/image/DOCX/audio work and keeps keyword-searchable derived text fresh; semantic indexing is skipped and reported as disabled. Audio without the selected local model is not queued and does not become a durable failure.
 
+Large-workload preflight consumes the same authoritative content-hash diff.
+Added and modified indexable sources count; hash-reused renames and no-op scans
+do not. Reconcile invalidates every known-stale modified row before publishing
+an awaiting or paused state. Invalidation failure or unavailable durable state
+falls back to normal indexing; it must never expose an unsafe or invisible
+pause. Known current prepared text contributes its byte size through bounded,
+asynchronous metadata reads, then the same format-specific completion,
+extractable-text, transcript-manifest, source-hash, and compatibility checks
+used by reconcile; mtime alone is never preparation truth. Preflight must not
+synchronously stat a large folder on the Node thread. Resume is serialized behind older folder reconcile
+work and clears the decision inside that queue entry. While paused, an upsert retains a row whose source hash is unchanged and
+invalidates a row whose hash changed. Conversion completion follows the same
+rule. Semantic retrieval additionally excludes daemon-pending identities for
+paused folders, so even a failed delete cannot publish known-stale evidence.
+Explicit Start indexing is the only operation that clears a durable pause.
+Every valid new convertible source counts before preparation exists; only its
+known-text byte contribution waits for a current validated representation.
+
 External writes are not immediately searchable. Editors, Git, cloud sync, terminal commands, and external Agents change the filesystem outside StashBase's write path. They become searchable after reconcile. StashBase-owned file helper writes should schedule or perform index maintenance as part of the write when possible.
 
 ---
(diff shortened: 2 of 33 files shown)


liliu-z:
Thanks for the thorough work here. The diagnosis, server-owned state model, regression coverage, and documentation are all strong, and the overall direction makes sense to me.

I reviewed the merge result against the latest `main`. The three-platform CI is green, and the relevant typecheck, semantic-state tests, renderer tests, Vite build, and server build pass locally.

I did find a few follow-up correctness gaps around the new lifecycle boundary:

- the pause gate is not active while stale rows are being invalidated, so a fast convertible-source completion can still upsert before the durable decision is published;
- `renameFile` / `renamePathPrefix` can fall back to re-embedding while the folder is paused;
- folder scoping needs tightening: paused writes should use the longest bound-folder owner, and renderer semantic state must be cleared when switching folders;
- workload estimation currently counts non-extractable text sources and measures audio transcript JSON rather than the Markdown that is actually embedded; and
- the two-worker validation limit is reached only after up to sixteen full DOCX/audio artifacts have already been read into memory.

These look fixable without changing the core approach. I’m going to merge this so the main state model and test harness land, then follow up with a maintainer patch for the lifecycle/resource issues above. Thanks again — the breadth of the implementation made the remaining trade-offs much easier to identify.

PR #157, opened 2026-08-12
feat(agent): render offline math in replies

### LINKED ISSUE #156: feat(agent-panel): render offline LaTeX math in responses
## Problem

Assistant responses in the built-in Agent Panel render LaTeX source literally instead of typesetting it. A response that should show a readable equation currently exposes commands such as `\text{}`, `\frac{}{}`, `\boxed{}`, and `\approx` in the transcript.

This is common in research, finance, engineering, and scientific conversations, where Claude or Codex naturally emits display math. The shared response renderer means the gap affects both runtimes, live turns, and resumed history.

## User impact

- Formula-heavy answers are difficult to read and easy to misinterpret.
- Long display equations wrap as raw source across the narrow side panel.
- Users get inconsistent behavior: ordinary Markdown documents support offline KaTeX after #28, while Agent responses do not.
- Asking the Agent to restate the answer without LaTeX is an unreliable workaround and loses mathematical structure.

### App-level reproduction

1. Open a library folder and start either a Claude or Codex chat.
2. Send this prompt:

   ```text
   Explain the following valuation equation and include it as display math.
   Do not put the equation in a code fence.

   \[
   \text{Present value}
   =
   \frac{\text{future EPS}\times\text{future P/E}}
        {1+\text{required return}}
   \]
   ```

3. Wait for the assistant response.

### Deterministic renderer reproduction

Render the shared Agent Markdown component with the same source:

```tsx
renderToStaticMarkup(
  <AgentMarkdown
    markdown={String.raw`
\[
\boxed{\$328.57 \approx \frac{\$13.26 \times 28}{1 + 13\%}}
\]
`}
    onOpenArtifact={() => {}}
  />,
)
```

## Actual behavior

The transcript displays literal source similar to:

```text
[ \boxed{\$328.57 \approx \frac{\$13.26 \times 28}{1 + 13\%}} ]
```

The Markdown escape handling may consume the backslashes on `\[` and `\]`, while the inner TeX commands remain visible.

## Expected behavior

The equation is typeset as accessible, readable math inside the assistant response. Inline math remains inline; display math occupies its own block and scrolls horizontally inside the response when it is wider than the available Agent-panel width.

Copy Reply must continue copying the original Markdown/LaTeX source rather than a lossy text extraction of the rendered formula.

## Confirmed current implementation

The shared renderer is [`web-src/src/components/agent/AgentMarkdown.tsx`](https://github.com/liliu-z/stashbase/blob/main/web-src/src/components/agent/AgentMarkdown.tsx). It currently configures `react-markdown` with `remark-gfm` only:

- no math syntax plugin creates math nodes;
- no KaTeX rehype renderer converts those nodes to React output; and
- the Agent surface does not own KaTeX styles/fonts when Chat is open without a Markdown document.

Focused coverage in [`web-src/src/__tests__/agent-markdown.test.ts`](https://github.com/liliu-z/stashbase/blob/main/web-src/src/__tests__/agent-markdown.test.ts) verifies GFM, raw-HTML inertness, link policy, and remote-image blocking, but has no formula cases.

This is intentionally separate from #28. That completed issue added offline KaTeX to ordinary Markdown preview and explicitly left Agent-message Markdown unchanged. No Agent server, Claude/Codex adapter, session protocol, or MCP change should be necessary here.

### Supported input

Support the common delimiters produced by Agent runtimes:

- inline: `$...$` and `\(...\)`;
- display: `$$...$$` and `\[...\]`.

Delimiter handling must be Markdown-aware:

- inline code and fenced code blocks remain literal;
- escaped delimiters remain literal;
- ordinary currency text such as `The price is $328.57` must not consume unrelated prose as math;
- incomplete delimiters during streaming must not hide or corrupt the surrounding response; and
- malformed/unsupported TeX must degrade to bounded visible source or a non-destructive KaTeX error representation instead of crashing the transcript.

At minimum, cover `\text{}`, `\frac{}{}`, `\times`, `\approx`, `\boxed{}`, escaped currency symbols, percentages, Latin text, and CJK text inside `\text{}`.

### Offline rendering and layout

- Render through locally bundled KaTeX assets; do not use a CDN, runtime fetch, or remote fonts.
- Declare packages imported by Agent code as direct dependencies rather than relying on Milkdown/Crepe transitive dependencies.
- Load the required KaTeX CSS/fonts even when Chat is the primary workspace and no Markdown document has mounted.
- Keep inline math aligned with surrounding prose.
- Contain wide display math within `.agent-prose`; horizontal overflow belongs to the equation block, not the entire chat panel.
- Preserve light/dark readability and the current compact Agent-panel typography.
- Avoid a large eager renderer-chunk regression; keep the existing lazy Agent/renderer boundaries effective.

### Existing trust boundary

Math support must not weaken the current Agent-message rendering policy:

- keep raw HTML inert;
- do not introduce `rehype-raw` or render an HTML string;
- keep remote images blocked;
- keep non-HTTP(S), non-workspace links inert;
- keep workspace links routed through the existing folder-safe callback;
- use KaTeX's untrusted/default behavior and do not enable `trust`; and
- do not create any new renderer network surface.

## Suggested implementation path

1. Extend the `AgentMarkdown` unified pipeline with math parsing and KaTeX rendering, typically through direct `remark-math`, `rehype-katex`, and `katex` dependencies.
2. Add a small Agent-specific, syntax-aware delimiter extension for `\(...\)` and `\[...\]` if the selected `remark-math` version does not parse them. Do **not** globally regex-rewrite the raw Markdown string: that can mutate code fences, inline code, escaped examples, and partial streaming output.
3. Import/own the local KaTeX stylesheet from an Agent-reachable renderer boundary, then add narrow-panel overflow and spacing rules under `.agent-prose` in [`web-src/src/styles/chat.css`](https://github.com/liliu-z/stashbase/blob/main/web-src/src/styles/chat.css).
4. Preserve the original assistant block text as the source for Copy Reply and transcript persistence; formula rendering should remain a presentation transform only.
5. Add focused renderer tests before broad UI work. Keep the math pipeline in `AgentMarkdown` so Claude and Codex cannot drift into separate implementations.

Equivalent implementations are welcome if they meet the behavior, security, streaming, and bundle constraints above.

## Acceptance criteria

- [ ] Inline `$...$` and `\(...\)` render as inline KaTeX in Agent responses.
- [ ] Display `$$...$$` and `\[...\]` render as display KaTeX in Agent responses.
- [ ] The valuation example above renders correctly, including `\text{}`, `\frac{}{}`, `\boxed{}`, currency, percentages, and CJK text.
- [ ] The same output works for Claude and Codex, for live/streaming turns, and after restoring session history.
- [ ] Incomplete and invalid formulas never crash, blank, or permanently hide a response while it streams.
- [ ] Code spans, fenced code blocks, escaped delimiters, and ordinary currency prose remain literal.
- [ ] Wide display equations scroll within the response without widening or overflowing the Agent panel.
- [ ] KaTeX assets work fully offline in both chat-primary and document-side-panel layouts.
- [ ] Copy Reply returns the original Markdown/LaTeX source.
- [ ] Existing GFM tables/task lists, raw-HTML handling, remote-image blocking, and link policy remain regression-covered.
- [ ] No server, session protocol, MCP, indexing, or filesystem-authority behavior changes.
- [ ] Renderer chunk-budget checks, type-check, focused tests, and the production renderer build pass.
- [ ] The Agent-panel product and review contracts are updated in the same PR.

## Regression coverage to add

Add focused cases to `web-src/src/__tests__/agent-markdown.test.ts` (and a browser-level layout test where static rendering is insufficient) for:

1. inline and display forms of both delimiter families;
2. the full valuation formula, including CJK `\text{}`;
3. literal currency prose and escaped dollar signs;
4. inline code and fenced code containing formula-like source;
5. incomplete opening/closing delimiters representing intermediate streaming states;
6. invalid and unsupported TeX;
7. raw HTML, unsafe links, and remote images beside valid math;
8. original-source Copy Reply behavior;
9. restored-history rendering; and
10. a wide display equation in a constrained Agent-panel width.

Tests should assert observable output/safety behavior rather than a particular plugin's internal AST shape.

- `web-src/src/components/agent/AgentMarkdown.tsx`
- `web-src/src/components/agent/AgentMessages.tsx`
- `web-src/src/__tests__/agent-markdown.test.ts`
- `web-src/src/styles/chat.css`
- `package.json` / `pnpm-lock.yaml`
- `scripts/check-renderer-chunks.mjs`
- `design-docs/design/agent-panel.md`
- `code-review/agent-panel.md`
- `README.md` if its external Agent-formatting summary is affected

## Validation

Run at minimum:

```bash
pnpm typecheck
pnpm test:renderer
pnpm test:renderer-chunks
```

Also run the new targeted Agent Markdown tests directly while iterating.

## Out of scope

- Changing ordinary Markdown document rendering already covered by #28.
- A visual equation editor, equation numbering, cross-references, or user-defined persistent macros.
- MathJax or any remote rendering service.
- Changes to Agent prompting, model output, transport, tools, permissions, or transcript storage.
- Enabling arbitrary raw HTML in assistant messages.

akira-in-tech:
Closes #156

## Summary

- render inline and display math in the shared Agent Markdown surface with locally bundled KaTeX
- normalize `\\(...\\)` and `\\[...\\]` through a Markdown-aware scanner while preserving code, escapes, incomplete streams, and ordinary currency prose
- keep the existing raw HTML, remote image, link, and KaTeX trust boundaries intact
- contain wide display equations inside the response and preserve original assistant source for Copy Reply and history
- document the Agent-panel behavior and review contract

## Validation

- `pnpm typecheck`
- `pnpm test:renderer` (221 passed)
- targeted Agent Markdown tests (8 passed)
- `pnpm build:web`
- renderer chunk budget: 12 dynamic entries; initial static JS 412613 / 425984 bytes

The production build includes local KaTeX font assets and does not add a renderer network dependency.


liliu-z:
Thanks for the thoughtful first contribution — the shared renderer approach, offline KaTeX path, and trust-boundary work are all heading in the right direction. I found a few integration follow-ups around currency-range false positives, theme-safe error styling, the unused Lucide dependency, lazy ChatPane weight, and browser-level copy/history/overflow coverage. I’ll take care of those maintainer-side while bringing the branch up to current main, so you don’t need to rework the PR before it can land.

akira-in-tech:
Thanks for merging, and for offering to handle the follow-ups — really appreciate it. Happy to pick any of those back up myself if it's easier on your end, just let me know.

PR #159, opened 2026-08-12
feat(mcp): let search_library do keyword (exact) search, not only semantic

### LINKED ISSUE #158: feat(mcp): let search_library do keyword (exact) search, not only semantic
## Problem

The retrieval layer supports two modes, `semantic` and `keyword`, and the in-app Search panel exposes both through the `≈` / `=` toggle. But `LibraryOperations.search` hardcodes `mode: 'semantic'` (`server/library-operations/index.ts`), so the MCP `search_library` tool can only run semantic search. Agents cannot run the exact, no-embedding keyword search that people can.

Two concrete consequences:

1. **No exact matching.** Hybrid semantic search blurs literal tokens. An agent looking for an exact identifier, error code, config key, or quoted phrase cannot ask for a literal match the way a person can with `=` mode.
2. **No search at all without an embedding key.** Semantic mode returns `unavailable` (HTTP 412, `EMBEDDER_KEY_REQUIRED`) when no key is configured, so `search_library` fails. Yet the app deliberately keeps keyword search working in exactly that situation ("Keyword search works without embeddings"). An MCP client connected to a keyless StashBase currently cannot search the library at all.

## What to build

- Add an optional `mode` (`semantic` | `keyword`, default `semantic`) to `search_library` and thread it through `LibraryOperations.search` to the existing retrieval seam, which already implements both modes and returns a unified `SourceEvidence` list.
- Keyword mode reuses the keyword options already on the retrieval query: `case_strict`, `whole_word`, plus the `path_prefix` and `types` filters that already exist. Keyword retrieval requires a folder scope, so the tool should require `folder` (or `path_prefix`) in keyword mode and return a clear error otherwise.
- Return the same hit shape regardless of mode (keyword hits map to a source path plus matched-line evidence), and surface the existing `truncated` / `partial` signal so an agent knows when keyword results were capped.
- Update the tool description and the docs (README MCP section, `architecture.md`, `design/search.md`) to state that MCP search offers the same two modes the app does.

## Out of scope

- Changing ranking or the hybrid semantic algorithm.
- A new result format; keyword hits reuse the existing evidence/hit shape.
- Any UI change.

## Acceptance criteria

- [ ] `search_library` accepts `mode: "keyword"` and returns exact-match results without an embedding key.
- [ ] Keyword mode honors `case_strict`, `whole_word`, `path_prefix`, and `types`, and requires a folder scope with a clear error when it is missing.
- [ ] Semantic behavior is unchanged when `mode` is omitted.
- [ ] Result identity stays a user-visible source file, and derived-note hiding is preserved in both modes.
- [ ] Typecheck and the MCP and retrieval test suites pass.

jashkarangiya:
Closes #158.

## Problem

The retrieval seam already supports both `semantic` and `keyword` modes, and the in-app Search panel exposes both through the `≈` / `=` toggle. But `LibraryOperations.search` hardcoded `mode: 'semantic'`, so the MCP `search_library` tool could only run semantic search. Agents could not do exact matching, and could not search the library at all when no embedding key was configured, even though the app keeps keyword search working in exactly that case.

## What changed

- `search_library` gains an optional `mode` (`semantic` default, or `keyword`), plus `case_strict` and `whole_word`. These thread through `operations.search` to the existing retrieval path, which already implements both modes and returns a unified source-evidence list.
- Keyword mode reuses the existing `path_prefix` and `types` filters, requires a folder scope (clear 400 otherwise, since a bare prefix would silently widen), and returns the same flat hit shape as semantic (path, matched line content, line number). Truncated keyword results surface a `truncated` flag.
- Semantic behavior is unchanged when `mode` is omitted.

This follows the same plumbing pattern as the existing `types` filter, so it stays contained: operations interface and impl, the `/api/library/search` route, the MCP HTTP adapter, and the tool schema/description, plus docs.

## Verification

- Typecheck, full build (web + server + mcp), and the library-operations, retrieval, and mcp test suites pass. New operations tests cover the folder-scope guard and truncated surfacing.
- Verified end-to-end against a real member folder with an invalid embedding key (so semantic embedding is unavailable): `mode: "keyword"` returned real hits (file path, line number, matched content) with no embedding call, a library-wide keyword request without a folder returned 400, and `case_strict` was honored (a case-mismatched query returned zero hits).

## Notes

- Keyword hits reuse the existing `SourceEvidence` to hit conversion, so agents get one consistent result shape across modes.
- The app's grouped keyword path (`operations.keywordSearch`) is untouched; this exposes the flat, agent-friendly shape through `search_library`.


server/library-operations/index.ts
```diff
@@ -115,17 +122,24 @@ export function createLibraryOperations(
...
 
-    async search({ query, topK = 8, folder, pathPrefix, types }) {
+    async search({ query, topK = 8, folder, pathPrefix, types, mode = 'semantic', caseStrict, wholeWord }) {
       const trimmedQuery = query.trim();
       if (!trimmedQuery) throw routeError('query required', 400);
       const scope = normalizeLibrarySearchScope(folder, pathPrefix);
+      // Keyword retrieval walks a folder subtree, so it needs a scope; a
+      // bare prefix without a folder would silently widen to the library.
+      if (mode === 'keyword' && !scope.folderRoot) {
+        throw routeError('keyword search requires a folder scope; pass `folder` or `path_prefix`', 400);
```
liliu-z, line 134: A valid keyword request with only path_prefix is rejected here: normalizeLibrarySearchScope validates the prefix but does not return its owning member as folderRoot. This contradicts the issue and the error text, which say either folder or path_prefix is sufficient. Please derive the owning member root from the normalized prefix, or route this mode through the existing library-wide keyword operation.

mcp/library-server.ts
```diff
@@ -103,21 +103,26 @@ export function createLibraryMcpServer(opts: LibraryMcpServerOptions): Server {
...
       const types = args.types == null ? undefined : parsedTypes;
+      const mode = args.mode === 'keyword' ? 'keyword' : 'semantic';
+      const caseStrict = args.case_strict === true;
+      const wholeWord = args.whole_word === true;
       const k = Math.max(
         1,
         Math.min(MAX_TOP_K, Math.floor(typeof args.top_k === 'number' ? args.top_k : DEFAULT_TOP_K)),
       );
-      const searchResult = await operations.search({ query, topK: k, folder, pathPrefix, types });
+      const searchResult = await operations.search({ query, topK: k, folder, pathPrefix, types, mode, caseStrict, wholeWord });
```
liliu-z, line 123: topK is forwarded, but Retrieval keyword mode never consumes it, so a response can report top_k 8 while returning far more than eight hits. The new mode should preserve the existing bounded-result contract: cap keyword evidence and mark the response truncated when that cap is applied.

server/routes/library-files.ts
```diff
@@ -56,12 +56,16 @@ export function mount(app: express.Express, operations: LibraryOperations = crea
           code: 'INVALID_SEARCH_TYPES',
         });
       }
+      const mode = req.body?.mode === 'keyword' ? 'keyword' : 'semantic';
```
liliu-z: This silently maps every unsupported mode to semantic. The default should apply only when mode is omitted; malformed values should return 400 rather than unexpectedly invoking embedding-dependent search. The MCP handler has the same coercion, so a shared validating parser would keep both adapters aligned.

server/library-operations/index.test.ts
```diff
@@ -55,6 +55,37 @@ test('Library Operations forwards file-type filters to Retrieval', async () => {
   assert.deepEqual(searchInput?.types, ['pdf', 'docx']);
 });
 
+test('Library Operations requires a folder scope for keyword search', async () => {
```
liliu-z, line 58: These additions cover rejection and generic truncation, but not the feature success path. Please add coverage proving keyword mode reaches Retrieval without embedding availability, forwards case_strict, whole_word, types, and scope, honors the result bound, and crosses the MCP or HTTP adapter with the expected flat response.

PR #177, opened 2026-08-16
feat(xlsx): add safe workbook support

### LINKED ISSUE #165: feat(library): support XLSX as a read-only searchable workbook
## Summary

Add `.xlsx` as a first-class **read-only convertible source**. Users should be able to open a workbook in StashBase, navigate its worksheets, copy visible cells, search its contents, and let Agents read a source-attributed textual representation.

StashBase must not edit or save workbook content. The original OOXML package remains the source of truth and can still be changed by an external spreadsheet application.

This is intentionally separate from #127, which only made unsupported files understandable. It is also separate from the editable CSV work in #164: XLSX is a compound packaged document with formulas, styles, relationships, drawings, and application-specific fidelity requirements.

## Problem

XLSX files are currently hidden as unsupported files. A user cannot inspect a workbook next to related notes and documents, find a cell through StashBase search, or give an Agent useful workbook context.

Treating XLSX like CSV would create a misleading editor. Rewriting an OOXML package safely requires preserving formulas, styles, tables, drawings, external relationships, metadata, and workbook semantics that StashBase does not own.

## Product decision

XLSX belongs to the same product class as PDF and DOCX: the source is visible and searchable, but StashBase is a viewer, not the owning editor.

- Scope is modern `.xlsx` only, case-insensitive.
- Do not include `.xls`, `.xlsm`, `.xlsb`, password-protected/encrypted workbooks, or macro execution.
- The source workbook remains in its original folder.
- Direct preview and durable preparation are independent success paths.
- Prepared text is hidden AppData-derived data, never a sibling note and never the displayed source identity.
- Normal file operations such as import, rename, move, delete, and external refresh remain supported. Workbook-content writes from the UI or Agent/MCP are not.
- Office temporary/lock files such as `~$...` and `.~...` must not surface as user documents or preparation work.

### Direct preview: `@extend-ai/react-xlsx`

Use [`@extend-ai/react-xlsx`](https://github.com/extend-hq/react-xlsx) as the preferred renderer viewer. It is MIT-licensed, supports React, renders worksheets with a worker/WASM-backed engine, and includes worksheet navigation, selection, zoom, charts/images, and large-workbook safeguards.

Only its read-only viewing surface belongs in this issue. Disable or omit mutation/export affordances that imply StashBase can save workbook changes.

This dependency is comparatively large and includes WASM. Before integrating it, complete a focused compatibility and packaging spike covering:

- the repository's current React version;
- Vite code splitting and Electron CSP;
- worker and WASM asset URLs in development and in every packaged platform;
- fully offline operation with no CDN dependency;
- cancellation and teardown when a tab or folder retires;
- initial renderer bundle impact;
- realistic memory use and startup time.

Do not merge the package before the spike proves both development and packaged-app loading.

### Parsing and durable extraction: prefer `@dukelib/sheets-wasm`

Prefer the same underlying [`@dukelib/sheets-wasm`](https://www.npmjs.com/package/@dukelib/sheets-wasm) engine for both renderer parsing and server-side durable extraction if its Node worker and packaging behavior pass the spike. Reusing one parser reduces preview/index disagreement.

If it cannot support bounded server-side extraction, [ExcelJS](https://github.com/exceljs/exceljs) is the preferred MIT-licensed fallback candidate. Do not add a second parser silently: document the divergence, add shared fixtures that compare displayed and extracted cell content, and coordinate before adopting it.

Do not depend on a runtime CDN or introduce a non-permissive/commercial spreadsheet engine.

### Discovery and identity

- `.xlsx` files appear in the folder tree, Quick Open, recent-document flows, and source-aware actions.
- Add a single search category labeled **Spreadsheets** (internal naming may remain centralized) and map `.xlsx` to it.
- Tabs, previews, search results, citations, and Agent responses always identify the real `.xlsx` path, not a prepared artifact.

### Direct read-only preview

- Open the source directly through the asset boundary without waiting for preparation.
- Show worksheet tabs in source order and allow keyboard/mouse navigation, scrolling, zoom, cell/range selection, and copy.
- Render common values, dates, text, booleans, errors, merged cells, frozen panes, styles, images, and charts to the extent supported by the selected viewer.
- Formulas may display cached/calculated values already present in the file. StashBase must not promise recalculation or execute formulas.
- External links, data connections, embedded scripts, and remote resources are never followed or executed.
- The UI must clearly say **Read only** and expose no in-app save path.
- If the direct renderer fails but prepared text is current, a truthful textual fallback may be offered. Never present a stale or partial render as complete.
- Corrupt, encrypted, unsupported, oversized, or resource-exhausting workbooks receive a specific recoverable state rather than an empty viewer.

### Durable preparation, search, and Agent access

The preparation pipeline should create a deterministic, source-attributed textual representation under AppData.

At minimum, extracted content must preserve:

- workbook worksheet order and worksheet names;
- stable A1-style cell coordinates;
- displayed/cached cell values as text;
- formula source as inert, clearly distinguished text when available;
- meaningful row boundaries and enough local context for retrieval.

The representation may be stored as Markdown or another existing sanitized text artifact, but it must have the repository's normal source fingerprint/completion semantics. It is a cache, not an editable source and not a visible sibling document.

Required behavior:

- Preparation runs through the single daemon-owned queue with explicit limits, retries, failure states, and stale-generation rejection.
- Keyword and semantic search return the original workbook path and useful sheet/cell evidence.
- Agent/MCP reads return the current prepared representation with workbook, sheet, and cell identity. They must not expose the raw ZIP package as text or accept workbook-content edits.
- A changed source invalidates and reschedules preparation; stale prepared text is not silently served as current.
- Direct preview remains available when preparation is pending or failed, and preparation may succeed when the rich preview cannot render a feature.

### Performance and safety

XLSX is a ZIP-based container and must be treated as untrusted input.

- Enforce explicit compressed size, expanded size, worksheet count, cell count, image-byte, memory, and parsing-time bounds.
- Detect and reject ZIP bombs, malformed relationships, unsafe paths, and unsupported encryption without partial success claims.
- Parse off the renderer main thread.
- Lazy-load all XLSX renderer code and register its intended dynamic entry in `scripts/check-renderer-chunks.mjs`.
- Revoke object URLs and terminate workers when tabs/folders retire.
- Carry a source generation/version through async work and reject late results after refresh, rename, delete, or folder switch.
- No workbook content may initiate network access.

## Out of scope

- Editing cells or formulas, saving workbook content, recalculation, comments, collaboration, or tracked changes.
- `.xls`, `.xlsm`, `.xlsb`, password-protected/encrypted files, macros, Power Query, external data refresh, or VBA.
- Pixel-perfect parity with Microsoft Excel for every chart, font, drawing, or print layout.
- Converting XLSX into CSV or a user-visible Markdown note.
- Fixing unrelated existing file-conflict behavior.

## Acceptance criteria

- [ ] A simple `.xlsx` workbook opens in a read-only viewer without waiting for preparation.
- [ ] A multi-sheet workbook exposes worksheet tabs, navigation, selection/copy, zoom, and correct source identity.
- [ ] Common values, cached formula results, merged cells, frozen panes, images, and charts are covered by representative fixtures, with unsupported features disclosed truthfully.
- [ ] No viewer action or Agent/MCP call can save workbook-content changes.
- [ ] Current prepared text includes worksheet names, cell coordinates, and values, and supports keyword search, semantic search, and Agent reads attributed to the original `.xlsx`.
- [ ] Preview and preparation can succeed or fail independently.
- [ ] Source modification invalidates stale prepared data and refreshes an open preview without accepting late prior-generation results.
- [ ] Office lock/temp files are ignored.
- [ ] Corrupt, encrypted, oversized, and adversarial ZIP inputs fail within documented resource bounds.
- [ ] Renderer code, workers, and WASM are lazy-loaded, work offline in packaged builds, and remain within the approved chunk budget.
- [ ] Existing PDF, DOCX, image, audio, Markdown, HTML, and JSON behavior does not regress.

## Suggested implementation areas

- Central shared format detection and spreadsheet search-category mapping.
- Asset delivery and a lazy renderer document component/worker boundary.
- Preparation scheduling, durable artifact ownership, completion markers, cleanup, and stale-generation handling.
- Indexability, keyword/semantic retrieval, citations, and Agent/MCP prepared-source reads.
- Library file listing and Office temporary-file exclusion.
- Renderer chunk registration, packaged WASM assets, CSP, and lifecycle cleanup.

Treat these as entry points, not a requirement to preserve today's file layout. Keep format policy centralized and keep direct preview independent from preparation.

## Regression coverage

Add focused tests for:

- extension and temporary-file detection;
- viewer format dispatch and Spreadsheets category filtering;
- a minimal workbook, multiple sheets, formulas with cached values, merged/frozen cells, and representative images/charts;
- direct-preview success while preparation is pending/failed, plus prepared fallback behavior;
- deterministic extraction with worksheet and A1 cell evidence;
- source refresh, preparation invalidation, cancellation, and stale-result rejection;
- Agent/MCP read-only access and source attribution;
- corrupt/encrypted input and ZIP/resource bounds;
- worker/WASM loading in development and packaged-app-compatible builds;
- lazy renderer chunks and teardown;
- end-to-end open, search, and reopen journeys.

## Documentation and validation

Update the affected product and engineering contracts in the same change, including `design-docs/design/documents.md`, `design-docs/design/preparation.md`, `design-docs/design/search.md`, `code-review/document-viewers.md`, `code-review/data-lifecycle.md`, `code-review/renderer-workspace.md`, and journey coverage where the behavior becomes Shipping.

At minimum, run:

- `pnpm typecheck`
- `pnpm build:web`
- `pnpm test:renderer`
- `pnpm test:renderer-chunks`
- `pnpm test:library-files`
- `pnpm test:conversion-scheduler`
- `pnpm test:retrieval`
- `pnpm test:python` if daemon extraction/indexing changes
- `pnpm test:mcp` if the Agent/MCP read path changes
- `pnpm test:e2e:check-focus` for E2E changes
- `pnpm test:e2e:smoke`
- `pnpm test:e2e:functional`
- `pnpm test:docs` for contract/journey updates

For covered composition changes, update Linux visual baselines only through the repository's **Generate visual baselines** workflow.

Swapnanilmanna1701:
## Summary

Adds first-class, read-only XLSX support across preview, preparation, search, Agent, and MCP surfaces while preserving StashBase's visible-source identity and local-only trust boundary.

Closes #165.

## What changed

- adds a lazy-loaded workbook viewer with worksheet tabs, zoom, frozen and merged cells, images, charts, internal worksheet navigation, range selection, and clipboard copy;
- keeps workbooks read-only and displays cached formula results without recalculating formulas;
- makes external cell, image, and shape hyperlinks inert while preserving internal worksheet links;
- adds bounded server-side XLSX-to-Markdown preparation for keyword search, semantic indexing, Agent context, and MCP reads;
- keeps derived Markdown in AppData and attributes all results and reads to the original `.xlsx` source;
- hides Office lock files such as `~$workbook.xlsx` and `.~workbook.xlsx` from listing and preparation;
- adds spreadsheet type filtering and XLSX source handling throughout renderer, retrieval, Agent, and MCP contracts;
- updates the affected design documents and engineering review contracts.

## Safety and robustness

The XLSX container is inspected before parsing and rejects malformed, encrypted, ZIP64, unsafe-path, oversized-compressed, oversized-expanded, excessive-worksheet, and excessive-image inputs. Extraction separately bounds populated cells, scanned grid slots, generated output, cancellation, and worker runtime.

Prepared output is published atomically and bound to device, inode, size, mtime, and ctime. Timestamp-preserving replacement and source changes during indexing retire stale generations instead of publishing stale evidence.

The pinned `@extend-ai/react-xlsx` patch disables formula recalculation in both controller and worker paths and removes external `window.open()` behavior. The renderer build fails if either protection disappears during a dependency update.

Direct workbook preview and searchable preparation remain independent: a preparation failure reports unavailable searchable text without preventing the original workbook from opening.

## Regression coverage

- XLSX format detection and Office temporary-file exclusion;
- extraction order, worksheet names, A1 coordinates, cached values, and inert formulas;
- malformed, encrypted, ZIP64, unsafe-path, expanded-size, worksheet, cell, grid, image, output, cancellation, timeout, and source-size limits;
- timestamp-preserving replacement and rejection of late indexing generations;
- keyword and semantic retrieval with original-source attribution;
- Agent and MCP reads through bounded derived content;
- Electron coverage for merged/frozen cells, decoded embedded images, chart rendering, cached formulas, selection/copy, inert external links, internal worksheet navigation, exact-search reopen, and preview/preparation failure independence.

## Validation

- `pnpm typecheck`
- `pnpm build:web`
- focused XLSX and retrieval tests: 16 passed
- `pnpm test:conversion-scheduler`: 143 passed
- `pnpm test:library-files`: 10 passed
- focused XLSX Electron journeys: 2 passed
- `pnpm test:e2e:smoke`: 5 passed
- `pnpm test:e2e:check-focus`
- `pnpm test:docs`
- `git diff --check`

Visual baseline generation was not performed locally because repository policy requires intentional XLSX composition baselines to be generated on Linux through the visual-baseline workflow.


PPRAMANIK62 (changes requested):
Reviewed at merge-base `6de0b94` → `b85a412` (45 files, +1340/−47) against #165, on two axes: **Standards** (repo's documented contracts + a code-smell baseline) and **Spec** (faithfulness to the issue).

Requesting changes primarily on the merge gate in #165 (`Do not merge the package before the spike`) and the canonical capability matrix, plus a few correctness items in the safety preflight and preview lifecycle.

---

## Spec

### Missing / partial

**Packaging spike (merge gate).** #165: *"Do not merge the package before the spike proves both development and packaged-app loading"* — covering *"worker and WASM asset URLs in development and in every packaged platform"*, memory, and startup time. No spike artifact exists, and the PR's Validation section lists only `build:web` and dev-mode Electron E2E; there is no packaged-build evidence. The acceptance bullet *"work offline in packaged builds"* is therefore unverified. `XlsxPreview.tsx:11` (`?url` WASM import) and the library's internal worker URL are exactly what the spike was meant to prove.

**Memory bound.** #165: *"Enforce explicit compressed size, expanded size, worksheet count, cell count, image-byte, **memory**, and parsing-time bounds."* `server/xlsx.ts:12-21` carries every bound except memory, and the `Worker` (~`server/xlsx.ts:160`) is constructed with no `resourceLimits`.

**Teardown.** #165: *"Revoke object URLs and terminate workers when tabs/folders retire."* `XlsxPreview.tsx` only aborts the fetch (`return () => controller.abort()`, ~line 59). There is no unmount disposal of the viewer controller/worker, and no `revokeObjectURL` for `image.src` (~line 186).

**Coverage gap.** #165 requires *"source refresh, preparation invalidation, cancellation, and stale-result rejection"* at the preview layer, and acceptance requires *"refreshes an open preview without accepting late prior-generation results."* `e2e/journeys/formats-media.spec.ts` has no external-modification-refresh case; only server-side generation retirement is covered (`server/xlsx.test.ts`, "retires a late generation").

### Scope creep

`server/conversion.ts:311,323` adds `ctimeMs` to `SourceSignature`/`sameSourceSignature` for **all** convertible formats. XLSX freshness is already independently bound by `sourceIdentity()` (`server/xlsx.ts`: dev/ino/size/mtimeNs/ctimeNs). This changes PDF/DOCX/image/audio reconversion triggering — ctime bumps on `chmod`, xattr writes, and rename — and bears on the acceptance bullet *"Existing PDF, DOCX, image, audio, … behavior does not regress."*

### Implemented but looks wrong

**ZIP-bomb detection.** #165: *"Detect and reject ZIP bombs."* `inspectXlsxContainer` in `server/xlsx.ts` sums `bytes.readUInt32LE(offset + 24)` — the *declared* central-directory uncompressed size — with no compression-ratio check and no bound on the bytes the WASM parser actually inflates. A bomb declaring small uncompressed sizes passes preflight. No test covers a real bomb; `server/xlsx.test.ts` only forces `expandedBytes: 3` against honest headers.

**Preview/index disagreement on hidden sheets.** The worker in `server/xlsx.ts` iterates `0..workbook.sheetCount` unconditionally, while `XlsxPreview.tsx:~63` sets `showHiddenSheets: false`. Hidden-sheet cells become searchable and Agent-readable but are unopenable in the viewer — against *"Reusing one parser reduces preview/index disagreement"* and *"Never present a stale or partial render as complete."*

**Preview timeout arming.** `XlsxPreview.tsx:~68` returns early unless `controller.isLoading` is already true, so the 30 s parsing deadline never arms if the controller does not report loading. The *"resource-exhausting workbooks receive a specific recoverable state"* path can then hang on "Opening workbook…".

---

## Standards

### Documented-standard violations

**`design-docs/design/documents.md:57-66`** — the Format Capability Matrix, declared "the canonical product-facing account of Shipping format behavior", gains no XLSX row; the change adds a prose bullet at `:44` instead. `MAINTENANCE.md:57` requires "one canonical capability matrix in the narrowest owning Area", and `MAINTENANCE.md:94` makes matrix-vs-dispatch the review check.

**`code-review/mcp-access.md:41-43`** — still reads "`read_file` returns … current prepared PDF, DOCX, or media text", but `server/library-file-reader.ts:170` now returns `'xlsx-derived-md'` and `mcp/library-server.ts:390` advertises `spreadsheets`. `AGENTS.md`: "update … a review contract when its Interface, invariant, risk, or validation changes."

**`README.md:163-169` and `:190-194`** (format table, "Agents read" list) and **`design-docs/glossary.md:99`** (Preparation) enumerate PDF/DOCX/image/audio and were left stale. `AGENTS.md`: "Keep the affected docs current in the same change."

**`release-checklists/ui-sanity.md:90`** still says "PDF, DOCX, image, and audio fixtures" while `code-review/journey-coverage.md:159` now claims "PDF/OCR/DOCX/XLSX/media preparation … remain release evidence" — a contract asserting evidence the checklist does not contain.

**`web-src/src/features/documents/components/XlsxPreview.tsx:157-169`** — hand-rolled `role="tablist"`/`role="tab"` buttons with bespoke classes, and icon-only zoom buttons as `<Button … aria-label="Zoom out">−</Button>`. `code-review/renderer-styling.md` layers 5-6: primitives exist for `tabs`, and "`IconButton`/`Toolbar` (accessible name is a required prop)"; "new buttons … use these instead of bespoke classes".

**`server/routes/file-assets.ts:71-77`** — hand-rolls `res.status(422).json({ error: … })` though `sendError` is imported and used by every other branch in the same file (`:64`, `:98`).

### Code-smell observations (judgement calls, not blockers)

- **Repeated Switches** — `server/library-file-reader.ts:88-92, 106-110, 170, 224-228` plus `server/agent.ts:218` each extend the same `sourceFormat === 'docx' ? … : 'xlsx' ? … : 'audio' ? …` cascade. Five sites; one shared map would serve.
- **Duplicated Code / Middle Man** — `web-src/src/common/api/api.ts:256` `prepareXlsx` is byte-identical to `prepareDocx`/`prepareAudio` directly above it.
- **Duplicated Code** — `useDocumentActions.ts:68` `isXlsxName` re-inlines `/\.xlsx$/i` plus the `~$`/`.~` checks, in the same PR that centralizes exactly that into `server/format.ts:150 isOfficeTemporaryFile` and adds `SPREADSHEET_EXTENSIONS` to `shared/`.
- **Mysterious Name** — `server/xlsx.ts:194` `derivedXlsxIsComplete(_sourceAbs, …)`: the underscore signals unused, but the parameter is used at `:202`.
- **Mysterious Name** — `SEARCH_TYPE_CATEGORIES` gains `'spreadsheets'` among `pdf`/`docx`/`audio` format tokens, forcing a gloss in every doc and a category→`xlsx` map.
- **Divergent Change** — `scripts/check-renderer-chunks.mjs:43-53` now greps `node_modules/@extend-ai/react-xlsx` source. A chunk-size guard is doing dependency-patch verification; that protection deserves its own check.
- **Divergent Change** — `XlsxPreview.tsx` (202 lines) owns fetch, timeout, clipboard fallback (`:85-131`), zoom, and sheet chrome inline, while `code-review/document-viewers.md` records a "Preview-control Modules" convention (`usePdf*`, `findIframe.ts`) that `DocxPreview` follows.
- No `logger('xlsx')` (cf. `docx.ts:29`, `pdf.ts:37`) — worker failures surface only as thrown text; and `server/xlsx.ts:1` has a one-line header where sibling converters carry architectural docblocks.

---

The overall shape — independent preview and preparation paths, AppData-derived text attributed to the original `.xlsx`, container preflight before parsing, and the pinned patch for recalculation and `window.open()` — matches what #165 asked for. The items above are what stand between that and the acceptance checklist.

PPRAMANIK62:
Thanks for putting this together, and sorry it took me a while to get through it properly. There's real work here and a lot of it is right. The `ConversionSpec` wiring follows the existing PDF and DOCX lanes cleanly, format policy stays centralised in `shared/file-formats.ts`, and adding `'wasm-unsafe-eval'` to `CSP_PROD` only, with the reasoning written down, is exactly the right call. The docs came with the code instead of after it.

I'm going to close this one though, and I want to be straight about why.

The main thing is the pnpm patch against `@extend-ai/react-xlsx`'s built `dist/`. It edits bundled output at fixed offsets, and the only thing holding it in place is a string grep in `check-renderer-chunks.mjs` where two of the three checks are negative. If upstream renames `safeCalculate` or the minifier shifts, both checks pass and formula recalculation quietly comes back. So the guard can't tell "protection is present" from "the code changed shape". For a small dependency I'd live with that. For something this size, sitting on the read only guarantee we're making to users, I don't want a security boundary that depends on a vendored bundle keeping its identifiers.

Two other things pushed me the same way:

The ZIP preflight rejects legitimate files. `server/xlsx.ts:134` fails any entry over 100:1. I measured it. A realistic sheet with distinct values compresses about 5.7:1, but a sheet with one repeated shared string index, a status column or template rows, hits 411:1. That's an ordinary workbook and we reject the whole thing as suspicious. Worse, the same guard runs on the asset route now, so the user can't even open it. The absolute `expandedBytes` cap already covers the actual bomb risk.

And the preflight sits on the hot path. `inflateRawSync` over every entry, synchronously, on the Express event loop, on every asset fetch. Roughly 100ms of blocked loop at the 200MB bound plus a 200MB transient allocation that falls outside the worker's 256MB limit. Then the renderer decompresses the same file and `extractXlsxText` does it a third time.

None of that means the approach was wrong, it means the safety layer needs rethinking, and that's easier for me to do as one piece than as review rounds on top of a 48 file diff.

I'll raise my own PR for this. Where I reuse what you did, and I expect that to be a decent chunk, the format vocabulary work, the extraction shape, the fixtures, you'll be the author or co-author on those commits. I'll link it here when it's up so you can see what carried over.

Issue #164, opened 2026-08-13
feat(library): support CSV as a first-class editable tabular source


liliu-z:
## Summary

Add `.csv` as a first-class **direct-text structured-data** format. A CSV file should be visible, searchable, readable as a table, editable in a grid, and editable as raw source without being converted into a note or a hidden derived file.

This is intentionally separate from #127, which only made unsupported files understandable, and #149, which added JSON as a first-class structured-text source.

## Problem

CSV files are common interchange artifacts, but StashBase currently treats them as unsupported. That means a user cannot inspect a dataset in context, correct a cell, search it, or give an Agent useful access to it without leaving the workspace.

A naive implementation that parses the file and serializes the whole table on every save is not acceptable. It can silently change delimiters, quoting, line endings, leading zeros, empty rows, multiline fields, and other untouched source text.

## Product decision

CSV belongs to the same product class as JSON: the original file is readable source text and remains the canonical artifact, so StashBase may support both reading and writing it.

- Scope is `.csv` only, case-insensitive. Do not implicitly add TSV, XLSX, or other spreadsheet formats.
- CSV is structured data, not a note.
- No AppData-derived note or preview artifact is created.
- All cell values are strings. Do not infer dates, numbers, booleans, formulas, or schemas.
- Formula-like text such as `=SUM(A1:A2)` is displayed as text and is never executed.
- The existing shared file transaction, conflict, refresh, and close/save barriers remain authoritative.

### Grid: `@glideapps/glide-data-grid`

Use [Glide Data Grid](https://github.com/glideapps/glide-data-grid) as the preferred virtualized table UI. It is MIT-licensed and provides editing, selection, copy/paste, and large-table virtualization. Its [documentation](https://docs.grid.glideapps.com/) includes explicit clipboard integration.

Before adopting it, complete a small compatibility spike against this repository's current React version, Vite build, Electron runtime, keyboard model, and test environment. At issue creation time, the project's docs and published peer metadata are not fully aligned on React 19 support. Do not hide an incompatibility behind a peer-dependency override. If the spike fails, document the failure and coordinate before replacing the library.

### Parser: `csv-parse`

Use [`csv-parse`](https://csv.js.org/parse/) as the preferred parser. It is MIT-licensed, has a browser ESM distribution, supports streaming, and exposes [`raw`](https://csv.js.org/parse/options/raw/) record text and parse information.

[...]

PR #178, opened 2026-08-16
feat(library): support CSV as a first-class editable tabular source

krit22:
## What changed

- Added CSV format support across shared definitions, file format mapping, and desktop registry (`electron/main.cjs`).
- Implemented a dual-mode CSV document viewer (`CsvDocument.tsx` and `CsvTableView.tsx`) using `@glideapps/glide-data-grid` Canvas grid for Table mode, and CodeMirror 6 for raw Source mode.
- Fixed cell selection feedback loops and missing Glide Data Grid CSS modules by bundling and loading them locally.
- Added canvas zoom scaling via `Ctrl + Mouse Scroll Wheel` and keyboard shortcuts (`Ctrl + +` / `Ctrl + -` / `Ctrl + 0`) scaling headers, cells, text, and markers in sync.
- Handled RFC 4180 parsing, multiline quoted fields, BOM preservation, line endings (CRLF/LF), and ragged rows.
- Enhanced backend sync, keyword search, file save transactions, and library operations with protection against extension changes.
- Added friendly error parsing for index warnings, translating status codes (401, 403, 429) and raw JSON payloads into actionable user-facing descriptions.

## Why

To support CSV as a first-class editable tabular source (Issue #164). Previously, CSV files were not visible or editable as tables in StashBase. The Grid overlay editor required a DOM portal which was missing, causing cell editing to fail. Adding scale and visual enhancements improves tabular inspection. Surfacing raw error JSON from failed index operations degraded the user experience.

## User impact

Users can view and edit `.csv` files as structured spreadsheets. They can add or delete rows, zoom into cell contents using `Ctrl + Mouse Scroll Wheel`, search inside cells, or edit the raw source text. Error toasts from expired API keys or rate limits now explain what to do instead of displaying raw JSON payloads.

## Validation

- `pnpm typecheck` (Passed)
- `pnpm test:renderer` (290 passed)
- `pnpm test:conversion-scheduler` (136 passed)
- `pnpm test:docs` (Passed)
- `pnpm build:web` (Passed)

## Visuals

image

image

image


PPRAMANIK62:
Hey, thanks for the work you put into this, and sorry to land on this answer.

I'm going to close this one rather than run it through more review rounds. That isn't about effort, and it isn't about any single piece of the code. It's that the changes I'd want don't fit in a review thread. They're design level, and there are four of them.

The grid library needs the compatibility spike #164 asked for before we adopt it. We're on React 19 and glide declares 16, 17 and 18. The vendored `glide-grid.css` turns out to be the package's own stylesheets concatenated by hand, byte for byte, because we don't run the Linaria plugin. That's a real integration problem and it wants a decision, not a patch.

Canvas rendering puts the table outside Find, screen readers, Playwright and visual regression, which is why the E2E test can only exercise Source mode. So the cell edit and row and paste path, the part this PR writes into `file-transactions.md` as contract, ends up with no end to end coverage. That's structural, not an oversight.

`CsvDocument.tsx` is close to a clone of `JsonDocument.tsx`. Rather than review the clone, I'd rather pull the shared dual mode viewer out first and build CSV on top of it.

And the parse model re-analyzes the whole file three times per cell edit and once per pasted cell. On a 300 KB file a 20x5 paste blocks for about 700 ms. That wants an incremental model or a worker, which is what #164 asked for.

Any one of those I'd happily iterate on with you. All four at once means rewriting most of the 2350 lines through comments, and the branch is already conflicting against `main` after #183 moved the renderer tree. Closing and restarting is cleaner for both of us.

What I'm taking with me, because it's good:

The format plumbing is right and I'm keeping it close to as written. `shared/file-formats.ts`, `server/format.ts`, the `data` search category mapping, the file glyph, the state wiring.

The source span patcher is the correct idea. Patch the smallest range, never serialize the whole document. I'll rebuild the internals but the approach and several of the operations come from your work.

Your unit tests in `csv-source-model.test.ts` and the `table-data.csv` journey fixture with the BOM and CRLF are carrying over more or less intact.

I'll open the replacement PR shortly. Every commit that takes something from here gets a `Co-authored-by` trailer with your name and email, so it lands in your contribution history properly. If you'd rather I credit it differently, say so and I'll do that instead.

For what it's worth, #164 was a hard issue with an unusually opinionated spec, and the library it recommended turned out not to fit our React version. That's not a trap you set for yourself. I wrote up detailed notes while reviewing and I'm happy to post them or walk through them if you want to see the reasoning, or if you'd like to take a run at the shared viewer factoring yourself.

PR #179, opened 2026-08-16
fix(save): prevent stale editor saves from overwriting newer content

### LINKED ISSUE #175: Prevent stale editor saves from overwriting Agent or external changes
## Problem

Editable Markdown and JSON use versioned saves, but the renderer currently handles FILE_CHANGED by retrying without the original base version. A stale editor buffer can therefore overwrite newer content written by an Agent or another external tool.

This is an accepted v2 Known Gap, not a request to block the current release. Track the safer long-term interaction here.

## Desired experience

Provide an explicit conflict flow similar to VS Code:

- Never retry a conflicted save without its base version.
- Preserve both the dirty editor buffer and the newer disk source.
- Show a useful comparison between the two versions.
- Let the user reload disk content, keep and explicitly overwrite with the editor version, or merge before saving.
- Keep close and navigation behavior truthful while the conflict is unresolved.
- Use the same version authority for renderer, Agent, and MCP writes.

## Acceptance criteria

- Cover the sequence: editor reads V1, Agent writes V2 from V1, stale editor save from V1 is rejected.
- No automatic unversioned V3 can erase V2.
- Both V2 and the dirty editor buffer remain recoverable.
- The user can compare the versions and make an explicit resolution decision.
- Closing a tab or window cannot silently discard either side.
- Add deterministic transaction tests and a focused renderer or Journey regression.

## References

- design-docs/user-journeys.md: J03 Documents
- code-review/file-transactions.md: Known gap — renderer conflict recovery
- code-review/journey-coverage.md: J03 evidence

krit22:
Fixes #175.

## Summary

Prevents a stale renderer save from silently overwriting newer disk content. A `FILE_CHANGED` response now enters an explicit conflict flow instead of retrying without `baseVersion`.

## What changed

- Preserves both the current live editor buffer and one coherent newer disk content/version snapshot.
- Shows an accessible side-by-side comparison with explicit Reload, Merge, and Overwrite actions.
- Keeps navigation, folder switching, tab/window close, and reload behind the save barrier while a conflict is unresolved.
- Serializes resolution and close/discard decisions so a late unversioned overwrite cannot overtake another choice; failed decisions re-enable recovery.
- Saves merged drafts against the newer disk version and keeps edits typed during an in-flight save recoverable.
- Routes recovery reload through the renderer flush barrier, removes direct reload menu paths, and blocks Chromium reload accelerators including F5.
- Tracks real editor-change intent synchronously, so render-gap edits are saved without treating mount-time editor normalization as a user edit.
- Aligns the Documents design, file/window transaction contracts, J03 coverage, and packaged release check.

The unrelated transcription-sidecar download-mirror change was split into draft PR #180.

## Validation

- `pnpm typecheck`
- `pnpm build:web`
- `pnpm test:renderer` — 280 passed
- `pnpm test:electron` — 92 passed
- `pnpm test:docs`
- `pnpm test:e2e:check-focus`
- `env -u ELECTRON_RUN_AS_NODE pnpm test:electron:smoke`
- `env -u ELECTRON_RUN_AS_NODE pnpm test:e2e:smoke` — 5 passed
- `env -u ELECTRON_RUN_AS_NODE pnpm test:e2e:functional` — 30 passed
- focused conversion, library-file, and MCP suites

Final Standards and Spec reviews report no actionable findings.


krit22:
Fixes #175

Issue #198, opened 2026-08-25
feat(indexing): add per-project AI Index controls


liliu-z:
## Problem

AI Index authorization is currently application-wide. Once a hosted account or BYOK embedding source is active, every member folder in the library is eligible for semantic indexing and library-wide semantic retrieval.

Users may have a mixed library: some projects benefit from AI Index, while others should remain exact-search-only because of provider quota, cost, project sensitivity, or simple lack of value. Today there is no durable project-level control for that choice.

The existing controls do not fill this gap:

- **Skip AI Index for now** only dismisses setup for the current window context; it is not a server-owned indexing policy.
- The large-workload preflight from #124 can durably defer one folder, but only after its changed workload crosses the warning threshold. It is a cost/liveness safeguard, not a general project setting.
- Removing a key or signing out disables AI Index globally rather than for one project.

## Product decision

Add a durable **AI Index enabled/disabled policy per library project** while keeping the embedding source and credentials application-wide.

- Preserve current behavior by default: a project remains eligible for AI Index unless the user explicitly disables it.
- Expose the control from a project-owned surface, such as the folder menu or project settings, with clear current state and consequences.
- Disabling a project stops new embedding work for that folder, including automatic reconcile and backfill after embedding-source changes.
- Disabled projects do not contribute results to Similar Search in the app or through MCP. Exact Search and direct file access remain available.
- Keep existing compatible vectors locally so the choice is reversible. Re-enabling reconciles current disk state, reuses still-compatible vectors, and embeds only missing or changed content.
- A global unavailable embedding source still takes precedence. Restoring or switching the global source must not silently re-enable projects the user disabled.
- Keep project policy distinct from the large-workload awaiting/paused decision in #124. Re-enabling a project may still enter that preflight when the pending workload is large.

## State and correctness constraints

- Persist policy by normalized folder identity under the server-owned configuration/state boundary; renderer memory must not own the decision.
- Reconcile remains folder-explicit, and the per-project gate is checked before and between embedding calls so disabling a project stops a running batch at a safe boundary.
- Status must distinguish globally unavailable AI Index, project-disabled AI Index, large-workload pause, partial readiness, indexing, failure, and ready.
- Known-stale rows from a disabled project must not surface as semantic evidence.
- Library-wide semantic search must respect project policy without silently widening scope. A search explicitly scoped to a disabled project should explain that AI Index is disabled there.
[...]

PR #199, opened 2026-08-26
feat(indexing): support a local embedding model for AI Index

### LINKED ISSUE #195: feat(indexing): support a local embedding model for AI Index
## Problem

Semantic search (AI Index) currently requires an OpenAI-compatible embedding provider. `python/requirements.txt` documents this as intentional: "No local model, onnxruntime/tokenizers stay out of the default sidecar." Meanwhile OCR and audio transcription both already run fully on-device via bundled Whisper models.

This creates an inconsistency in the local-first story. Two of the three hard-to-read-format pipelines (OCR, transcription) work fully offline, but the core AI Index capability, which is arguably the main thing that sets StashBase apart from an agent just reading files directly, depends on a network call and either a hosted account or a BYOK cloud key.

## Goal

Offer a local embedding model as an alternative embedding source, so semantic search can work fully offline without an account or API key, the same way Whisper is already bundled for transcription.

## Acceptance criteria

- A local embedding option is selectable alongside the existing hosted/BYOK paths in Settings.
- Existing exact search and file workflows are unaffected when the local option is selected.
- Vectors produced by the local model are clearly distinguished from provider-based vectors, so switching providers doesn't silently mix
  incompatible embeddings in the same index.
- Documentation in `design-docs/` reflects the new local-first option.

VishakBaddur:
## Summary

AI Index currently requires an OpenAI-compatible embedding provider: `python/requirements.txt` documents this as intentional ("No local model, onnxruntime/tokenizers stay out of the default sidecar"), while OCR and audio transcription both already run fully on-device via bundled Whisper models. This adds a local, offline embedding option so AI Index can also work with no API key and no cloud call.

## What changed

- Adds `_LocalEmbedder` to `stashbase_daemon.py`: a thin wrapper around `mfs-cli`'s bundled ONNX embedding provider (`mfs.embedder.get_provider("onnx", ...)`, default model `gpahal/bge-m3-onnx-int8`, 1024-dim). No API key needed; the model downloads and caches locally on first use.

- `make_embedder("local", ...)` builds it; `bind_root`'s no-embedder gate (previously `if api_key and ...`) now also accepts `provider == "local"`, since local needs no key.

- Fixes a real correctness gap found while building this: `_collection_name()` only encoded dimension, not provider identity, a local model sharing a dimension with an OpenAI model (reachable via OpenAI's `dimensions` truncation parameter) would have silently mixed incompatible vector spaces into the same collection. Now provider-aware; `openai`/`openrouter`/`stashbase` keep the exact historical `vectors_openai_{dim}` name (verified unchanged, so existing indexed libraries aren't orphaned), `local` gets its own `vectors_local_{dim}` namespace.

- Adds `python/requirements-local-embed.txt` (`mfs-cli[onnx]`), following the exact precedent of `requirements-extract.txt` for other heavy, opt-in local dependencies. Kept out of the default `pnpm setup:python` flow.

- Out of scope: Settings UI for selecting "local" as a source, and any model-download UX (the ONNX model download is currently a blocking, unattended first-use cost, the app would want progress/consent UI for this before end users can reach it). This PR is the Python daemon capability only, matching the "the per-project policy should apply equally to local, hosted, and BYOK sources" framing from #198.

## Validation

- [x] `pnpm typecheck`
- [x] Focused tests for the affected behavior
- [ ] Renderer build or E2E coverage, when the change affects the UI

## UI and visual baselines

- [ ] This PR changes a rendered UI surface or visual state.
- [ ] This fork PR allows maintainer edits if a reviewed Linux visual baseline update is needed.
- [x] This PR does not change a visual surface, or existing visual baselines remain valid.

For an intentional UI change, maintainers generate and review Linux baselines after code review. Contributors do not need to run the baseline workflow or commit PNGs.

## Documentation

- [ ] Updated the relevant `design-docs/` and `code-review/` contract, when this changes documented behavior or invariants.
- [x] No documentation update is needed.

Fixes #195

### DIFF
--- a/python/stashbase_daemon.py
+++ b/python/stashbase_daemon.py
@@ -123,10 +123,10 @@ def make_embedder(provider: str = "openai", *, model=None, api_key=None, dimensi
 
     ``provider`` accepts OpenAI/OpenRouter and the loopback-only StashBase
     account broker (all rolled in-house — see `_OpenAIEmbedder`), or
-    ``"local"`` for a CPU-only, no-API-key embedder (see `_LocalEmbedder`).
+    ``"onnx"`` for a CPU-only, no-API-key embedder (see `_OnnxEmbedder`).
     """
-    if provider == "local":
-        return _LocalEmbedder(model=model)
+    if provider == "onnx":
+        return _OnnxEmbedder(model=model)
     if provider not in ("openai", "openrouter", "stashbase"):
         raise ValueError(f"unsupported embedder provider {provider!r}")
     if not api_key:
@@ -145,7 +145,7 @@ def make_embedder(provider: str = "openai", *, model=None, api_key=None, dimensi
     )
 
 
-class _LocalEmbedder:
+class _OnnxEmbedder:
     """CPU-only local embedding via mfs-cli's bundled ONNX provider — no API
     key, and no network needed once the model is cached locally.
     Wraps the raw `mfs.embedder` provider (which only exposes `.embed()`,
@@ -153,11 +153,16 @@ class _LocalEmbedder:
     `_collection_name()` reads it to keep local vectors in a separate
     collection from OpenAI's — the same dimension does not mean the same
     embedding space, and mixing them would corrupt search.
+    Named after the actual mfs provider it wraps ("onnx"), not the more
+    general "local" -- mfs also ships a *different* provider literally
+    named "local" (mfs.embedder.local.LocalEmbedding, sentence-transformers
+    based). Since vectors_<provider>_<dim> is disk-persisted and can't be
+    migrated later, "onnx" is the safer identifier to commit to.
     Requires the optional `mfs-cli[onnx]` extra
     (``pip install "mfs-cli[onnx]"``); raises mfs.embedder's own
     ImportError with install instructions if it isn't installed.
     """
-    provider = "local"
+    provider = "onnx"
     def __init__(self, *, model: str | None = None) -> None:
         from mfs.embedder import get_provider
         kwargs = {"model": model} if model else {}
@@ -836,7 +841,7 @@ class StashbaseStore:
         # A keyed bind attaches the embedder. A no-key bind still reopens an
         # existing collection for delete/list/status cleanup; it never creates
         # a new database merely because semantic indexing is disabled.
-        if (api_key or provider == "local") and self._embedder is None:
+        if (api_key or provider == "onnx") and self._embedder is None:
             embedder = make_embedder(
                 provider,
                 model=model,
--- a/python/stashbase_daemon.py
+++ b/python/stashbase_daemon.py
@@ -117,13 +117,15 @@ print(json.dumps({"event": "starting", "pid": os.getpid()}), flush=True)
 # an API key; the daemon may have zero embedders loaded at idle.
 
 def make_embedder(provider: str = "openai", *, model=None, api_key=None, dimension=None, base_url=None):
-    """Build an OpenAI-compatible embedding provider satisfying MFS's protocol
+    """Build an embedding provider satisfying MFS's protocol
     (`.embed(texts) -> list[list[float]]`, `.dimension`, `.model_name`).
 
-    Rolled in-house — see `_OpenAIEmbedder`. ``provider`` is accepted for
-    protocol compatibility and is intentionally limited to OpenAI/OpenRouter
-    plus the loopback-only StashBase account broker.
+    ``provider`` accepts OpenAI/OpenRouter and the loopback-only StashBase
+    account broker (all rolled in-house — see `_OpenAIEmbedder`), or
+    ``"local"`` for a CPU-only, no-API-key embedder (see `_LocalEmbedder`).
     """
+    if provider == "local":
+        return _LocalEmbedder(model=model)
     if provider not in ("openai", "openrouter", "stashbase"):
         raise ValueError(f"unsupported embedder provider {provider!r}")
     if not api_key:
@@ -140,6 +142,33 @@ def make_embedder(provider: str = "openai", *, model=None, api_key=None, dimensi
         dimension=dimension,
         base_url=base_url,
     )
+
+
+class _LocalEmbedder:
+    """CPU-only local embedding via mfs-cli's bundled ONNX provider — no API
+    key, and no network needed once the model is cached locally.
+    Wraps the raw `mfs.embedder` provider (which only exposes `.embed()`,
+    `.dimension`, `.model_name`) to add a `.provider` attribute, since
+    `_collection_name()` reads it to keep local vectors in a separate
+    collection from OpenAI's — the same dimension does not mean the same
+    embedding space, and mixing them would corrupt search.
+    Requires the optional `mfs-cli[onnx]` extra
+    (``pip install "mfs-cli[onnx]"``); raises mfs.embedder's own
+    ImportError with install instructions if it isn't installed.
+    """
+    provider = "local"
+    def __init__(self, *, model: str | None = None) -> None:
+        from mfs.embedder import get_provider
+        kwargs = {"model": model} if model else {}
+        self._inner = get_provider("onnx", **kwargs)
+    @property
+    def model_name(self) -> str:
+        return self._inner.model_name
+    @property
+    def dimension(self) -> int:
+        return self._inner.dimension
+    def embed(self, texts: list[str]) -> list[list[float]]:
+        return self._inner.embed(texts)
 
 
 class _OpenAIEmbedder:
@@ -586,12 +615,18 @@ def op_set_rules(svc: StashbaseStore, args: dict) -> dict:
     }
 
 
-def _collection_name(dim: int) -> str:
-    """The single collection's name. Encodes dim so a `text-embedding-3-
-    large` (3072) config wouldn't collide with the default small (1536);
-    the default keeps the historical `vectors_openai_1536` name so
-    already-indexed KBs aren't orphaned."""
-    return f"vectors_openai_{dim}"
+def _collection_name(dim: int, provider: str = "openai") -> str:
+    """The collection name for a given (dimension, provider) pair. Encodes
+    dim so a `text-embedding-3-large` (3072) config wouldn't collide with
+    the default small (1536); the default keeps the historical
+    `vectors_openai_1536` name so already-indexed KBs aren't orphaned.
+    Also encodes provider identity so a differently-sourced embedder (e.g.
+    a local model) can never silently share a collection with an
+    OpenAI-family one at the same dimension — same vector length does not
+    mean the same embedding space, and mixing them would corrupt search."""
+    if provider in ("openai", "openrouter", "stashbase"):
+        return f"vectors_openai_{dim}"
+    return f"vectors_{provider}_{dim}"
 
 
 def _norm_root(root: str) -> str:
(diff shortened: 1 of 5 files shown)


python/stashbase_daemon.py
```diff
@@ -688,7 +723,10 @@ def _ensure_store_for_dimension(self, dim: int, embedder=None):
         from mfs.store import MilvusStore
         from mfs.config import MilvusConfig
         os.environ["MFS_HOME"] = str(self._db_path.parent)
-        config = MilvusConfig(uri=str(self._db_path), collection_name=_collection_name(dim))
+        collection_provider = getattr(embedder, "provider", None) or "openai"
```
PPRAMANIK62: **Blocking. The provider-aware fix misses the attach path, so local vectors can still land in the OpenAI collection.**

`_collection_name` only gets consulted when a store is created. The early-return guard above at `python/stashbase_daemon.py:680-687` compares dimension alone, so when a store is already open it swaps in the new embedder and keeps the old collection:

```python
if self._store is not None:
    if embedder is not None:
        if self._dim != dim:          # provider never considered
            self.close_all(clear_bindings=False)
            return self._ensure_store_for_dimension(dim, embedder)
        self._embedder = embedder     # local embedder, OpenAI collection
    return self._store
```

What I got running this branch:

```
after no-key open : vectors_openai_1536  embedder=None
after local bind  : vectors_openai_1536  embedder.provider=local
```

`bind_root` reaches it. A no-key bind takes the `elif` reopen branch, which leaves `_store` set and `_embedder` at `None`. A following `provider="local"` bind then passes the `self._embedder is None` guard, needs no key, and attaches. Any time the local model's dimension matches the reopened one, you get the mixed vector spaces this PR is meant to prevent, including the 1024 case the description cites via OpenAI's `dimensions` truncation.

The guard has to compare the resolved collection name, not just `dim`.

python/stashbase_daemon.py
```diff
@@ -688,7 +723,10 @@ def _ensure_store_for_dimension(self, dim: int, embedder=None):
...
+        config = MilvusConfig(
+            uri=str(self._db_path), collection_name=_collection_name(dim, collection_provider),
```
PPRAMANIK62: **Blocking. A `vectors_local_*` collection can never be reopened without a credential.**

On the cleanup path `embedder` is `None`, so `collection_provider` falls back to `"openai"` and this always opens `vectors_openai_{dim}`. `code-review/data-lifecycle.md:112` states the invariant:

> An existing local collection may reopen without an embedding credential for list/delete cleanup.

For a library indexed with the local provider, that reopen quietly opens a different, empty collection. List and delete report zero rows, and the real rows are orphaned. The caller is `bind_root`'s `elif self._store is None and self._db_path.exists()` branch, which passes no embedder.

The cause is that nothing persists which provider an on-disk collection belongs to, so this path has no way to ask. Enumerating the collections already in the database and reopening whichever one exists would close this and the attach-path problem together. If that is too much for this PR, the limitation needs a Known Gap entry in `data-lifecycle.md` instead of quietly contradicting the contract.

python/stashbase_daemon.py
```diff
@@ -767,7 +805,10 @@ def bind_root(
             "provider": getattr(self._embedder, "provider", provider),
             "model": getattr(self._embedder, "model_name", None),
             "dim": self._dim,
-            "collection": _collection_name(self._dim) if self._dim else None,
+            "collection": (
+                _collection_name(self._dim, getattr(self._embedder, "provider", None) or "openai")
```
PPRAMANIK62: `bind_root` recomputes the collection name from the embedder's provider instead of reading it off the store it opened. In the attach-path case above the two disagree: this reports `vectors_local_1536` while the open store sits on `vectors_openai_1536`.

Reading `getattr(getattr(self._store, "_config", None), "collection_name", None)` reports what actually happened, and would have caught the attach bug rather than papering over it.

python/stashbase_daemon.py
```diff
@@ -142,6 +144,33 @@ def make_embedder(provider: str = "openai", *, model=None, api_key=None, dimensi
...
+    embedding space, and mixing them would corrupt search.
+    Requires the optional `mfs-cli[onnx]` extra
+    (``pip install "mfs-cli[onnx]"``); raises mfs.embedder's own
+    ImportError with install instructions if it isn't installed.
+    """
+    provider = "local"
+    def __init__(self, *, model: str | None = None) -> None:
+        from mfs.embedder import get_provider
+        kwargs = {"model": model} if model else {}
+        self._inner = get_provider("onnx", **kwargs)
```
PPRAMANIK62: This blocks the whole daemon on first use. `main()` is a serial `for line in sys.stdin` loop, and `OnnxEmbedding.__init__` downloads the model weights and runs a probe encode before it returns, all inside the `bind_folder` op. Search, status and delete stall for the duration, with no timeout. On a machine that is offline the first time, the download failure propagates straight out of `bind_root` and the folder just fails to bind.

Scoping the download UX out is fair. This part is different, though. It is a runtime property of the readiness barrier, not UI polish, so it needs a bounded failure mode here. A timeout and an error the Node side can report would be enough, so a slow or missing network degrades one bind instead of the whole sidecar.

python/stashbase_daemon.py
```diff
@@ -142,6 +144,33 @@ def make_embedder(provider: str = "openai", *, model=None, api_key=None, dimensi
     )
 
 
+class _LocalEmbedder:
```
PPRAMANIK62: Worth settling the name before merge, since `vectors_local_{dim}` goes to disk and cannot be migrated later.

StashBase's `"local"` maps to mfs's `onnx` provider, but mfs also has a separate provider actually named `local`, at `mfs.embedder.local.LocalEmbedding` behind the `mfs-cli[local]` extra. Adding that one later would want the same `vectors_local_*` namespace this claims. `"onnx"` or `"local-onnx"` is the safer thing to commit to now.

python/stashbase_daemon.py
```diff
@@ -117,13 +117,15 @@
...
     (`.embed(texts) -> list[list[float]]`, `.dimension`, `.model_name`).
 
-    Rolled in-house — see `_OpenAIEmbedder`. ``provider`` is accepted for
-    protocol compatibility and is intentionally limited to OpenAI/OpenRouter
-    plus the loopback-only StashBase account broker.
+    ``provider`` accepts OpenAI/OpenRouter and the loopback-only StashBase
+    account broker (all rolled in-house — see `_OpenAIEmbedder`), or
+    ``"local"`` for a CPU-only, no-API-key embedder (see `_LocalEmbedder`).
     """
+    if provider == "local":
```
PPRAMANIK62: `make_embedder` accepts `dimension` and `base_url` and then drops both for this provider. Ignoring `dimension` is correct, since the ONNX model's dimension is authoritative. But a caller that passes 1536 and gets a 1024 collection should get either a raise or a docstring line saying both are ignored.

PPRAMANIK62 (changes requested):
I applied the patch to a clean worktree at `main` and ran it. `node scripts/test-python.mjs` passes, 32 tests, with all three new ones included. The `mfs-cli` claims hold up: `mfs.embedder._PROVIDERS` maps `onnx` to `mfs.embedder.onnx.OnnxEmbedding`, the default model is `gpahal/bge-m3-onnx-int8`, and the install hint is exactly `pip install "mfs-cli[onnx]"`.

The `_collection_name` gap you found along the way is a real latent bug, and keeping the historical `vectors_openai_{dim}` spelling is the right call. `requirements-local-embed.txt` follows the `requirements-extract.txt` precedent correctly. No packaging wiring is needed, since the only other place that file gets referenced is macOS quarantine clearing.

Two things block this. The provider-aware collection fix only applies when a store is created, not when an embedder attaches to one that is already open, so the mixed vector spaces this PR is meant to prevent are still reachable. I left a repro inline. Separately, a `vectors_local_*` collection can never be reopened without a credential, which contradicts an invariant stated in `code-review/data-lifecycle.md`. Both come from the same gap. Nothing records which provider an on-disk collection belongs to, so one fix probably closes both.

### Minor

`stores()` at `python/stashbase_daemon.py:817` still hardcodes `f"openai_{self._dim}"` as the provider key. All 14 call sites discard it as `_pk`, so nothing breaks, but it is now wrong.

`"local"` cannot be reached end to end. The unions at `server/indexer.ts:19`, `server/mfs-daemon.ts:78`, and `shared/embedding.ts:13` exclude it, and `server/app-config.ts:321` requires a key. Scoping the Settings UI out is reasonable, but it means unit tests are the only evidence obtainable for this slice. Worth saying that outright in the description.

### Documentation

The description checks "No documentation update is needed", but this changes collection identity and no-key reopen behavior. Both are covered by `code-review/data-lifecycle.md`, and AGENTS.md requires that doc to move in the same change as the code. A Known Gap entry for the reopen limitation is the minimum.

VishakBaddur:
Thanks for the thorough review, both blocking issues were real, and you nailed the root cause: nothing on disk records which provider a collection belongs to. Fixed in f44352a.

Both blocking issues, fixed:

Attach path: _ensure_store_for_dimension() now compares the actual resolved collection name (_collection_name(dim, provider)) against what's really open (self._store._config.collection_name) before deciding to reuse vs reopen, instead of just comparing dim. Reproduced your exact repro (no-key open then local bind on the same instance) as a test against a real Milvus store before the fix, confirmed it reproduced, confirmed it doesn't anymore after.
Credential-less reopen: instead of defaulting to dimension or 1536 / "openai", it now connects a MilvusClient and calls list_collections() to see what's actually on disk. Exactly one match, opens it (dimension and provider parsed from the real name). Zero matches, no-op (this is actually more correct than before too, stores()'s own docstring says empty is expected before the first keyed bind, and the old code was eagerly creating a phantom collection, violating that). More than one match, raises clearly instead of guessing, since which one's "active" is genuinely ambiguous without a credential.

Also picked up your two non-blocking findings since they shared the same root cause:

bind_root's return dict reads the collection name directly off the real open store now instead of recomputing it from the embedder.
stores() no longer hardcodes "openai" as the provider key.

Added 3 regression tests reproducing each bug against a real Milvus store (not mocked) before confirming the fix works. pnpm test:python: 35/35 (32 previous + 3 new).

Still working through your other three points (the "local" vs mfs's actual local provider naming collision, the unbounded blocking model download with no timeout, and the data-lifecycle.md doc update), will push those separately.

VishakBaddur:
Following up on the other three, all pushed now.

Renamed _LocalEmbedder/"local" to _OnnxEmbedder/"onnx", matching your suggestion exactly, since mfs ships its own actual local provider and this is disk-persisted so worth settling before merge.
Added a bounded timeout around the ONNX model construction (background thread, 120s wall-clock cap, concurrent.futures). Looked into why before picking a number: huggingface_hub already fast-fails a dead network on its own (etag_timeout=10s), and downloads resume rather than restart (HTTP Range + .incomplete file), so a shorter bound is actually more robust than a longer one, keeps the daemon responsive sooner without losing progress on retry. Added a regression test that bounds a simulated hanging call and confirms it fails near the timeout instead of hanging.
Added a Known Gap entry in data-lifecycle.md for the one real remaining limitation the discovery fix doesn't close: credential-less cleanup genuinely isn't possible when a library has multiple historical collections, since nothing tracks which one's active. Matches the existing format in mcp-access.md.

pnpm test:python: 36/36, Windows and macOS CI both green.

